Issue #8, opened 2019-11-08
ResizeTool is not aware of MovementRestrictors


tortmayr:
see https://github.com/eclipsesource/graphical-lsp/issues/391

tortmayr:
Fixed with #27

planger:
@tortmayr Is this really fixed? I thought #27 only addressed the move-part but not the resize part? I think the resize handles don't respect the movement restrictors yet, do they?
I might be missing something though. Thanks!

tortmayr:
Yes this is fixed: ![prove](https://user-images.githubusercontent.com/2311075/74018914-03270680-4997-11ea-85cb-5a6499a16857.gif)

tortmayr:
Maybe you were thinking of the snapping of the resize tool?. This is still missing and tracked with #34

planger:
Ah ok, yes, I meant the issue that is tracked now in #34. :+1:

Issue #5, opened 2019-11-08
Adapt sprotty snapping feature for GLSP


tortmayr:
To use the new snapping facility of sprotty (eclipse/sprotty#87) in GLSP the ChangeBoundsToolFeedback needs to be adapted.

[migrated from https://github.com/eclipsesource/graphical-lsp/issues/257]

PR #27, opened 2020-01-20
#5 #8 Add snapping feature and rework movement restrictor

tortmayr:
- Add support for snapping feature of sprotty
- Rework NoOverlapMovementRestrictor
- Add feedback for element overlap
- FIx issue that  triggered a ChangeBoundsAction on click

### DIFF
--- a/src/base/model/update-model-command.ts
+++ b/src/base/model/update-model-command.ts
@@ -85,11 +85,10 @@ export class FeedbackAwareUpdateModelCommand extends UpdateModelCommand {
 
             const feedbackCommands = this.getFeedbackCommands(this.actionHandlerRegistry);
             feedbackCommands.forEach(command => command.execute(tempContext));
-
         }
+
         this.modelRootListeners.forEach(listener => listener.modelRootChanged(newRoot));
         return super.performUpdate(oldRoot, newRoot, context);
-
     }
 
     private getFeedbackCommands(registry: ActionHandlerRegistry): Command[] {
--- a/src/base/model/update-model-command.ts
+++ b/src/base/model/update-model-command.ts
@@ -29,8 +29,8 @@ import {
     TYPES
 } from "sprotty/lib";
 import { UpdateModelAction, UpdateModelCommand } from "sprotty/lib/features/update/update-model";
-import { IFeedbackActionDispatcher } from "src/features/tool-feedback/feedback-action-dispatcher";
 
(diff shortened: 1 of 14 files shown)


tortmayr:
Note:  Resize handles are currently not snapped (even if a snapper is bound). This will be addressed in a follow up.
resize

src/base/model/update-model-command.ts
```diff
@@ -80,10 +82,10 @@ export class FeedbackAwareUpdateModelCommand extends UpdateModelCommand {
...
                 syncer: context.syncer
             };
-            this.actionHandlerRegistryProvider().then(registry => {
-                const feedbackCommands = this.getFeedbackCommands(registry);
-                feedbackCommands.forEach(command => command.execute(tempContext));
-            });
+
+            const feedbackCommands = this.getFeedbackCommands(this.actionHandlerRegistry);
+            feedbackCommands.forEach(command => command.execute(tempContext));
+
```
planger: I feel this blank line is unnecessary.

src/features/change-bounds/movement-restrictor.ts
```diff
@@ -14,276 +14,76 @@
...
-            snappedBounds = {
-                x: target.bounds.x - 1 - element.bounds.width,
-                y: element.bounds.y,
-                width: element.bounds.width,
-                height: element.bounds.height
-            };
-        }
-        return snappedBounds;
-    }
+export function doOverlap(element1: SModelElement & BoundsAware, element2: SModelElement & BoundsAware) {
```
planger: Probably just a matter of taste, but I'd prefer `areOverlapping`, because `doOverlap` doesn't sound to me like a boolean-query ("is", "does", etc.) but rather like an operation with side-effect.

tortmayr:
Ad 1: We simply reused the `CenterGridSnapper` from Sprotty here but it should be no problem to bind a custom snapper that snaps on the top-left corner of an element.

Ad 2: I will address the minimum size behavior. As I already noted before the resize handles are currently not snapped on the grid. This will be addressed in a follow up PR

tortmayr:
I cannot reproduce this issue. For me the feedback is cleaned after an overlap.
I updated the counter PR https://github.com/eclipse-glsp/glsp-examples/pull/26 and elements should now snapp to the topleft

planger:
I tested it again and can provide more details for reproducing it:
I think you have to wait while dragging for the hover popup to appear and then drop on the overlapping position.
Thanks for the update of eclipse-glsp/glsp-examples#26. That works great now!!

Issue #31, opened 2020-02-05
Support popups for validation markers


jfaltermeier:
When hovering over validation markers a popup with the issues' message should be displayed. 

![Selection_913](https://user-images.githubusercontent.com/5889696/74037176-0ed9f380-49be-11ea-8254-34ddd8c27ddb.png)

PR #35, opened 2020-02-07
Support popups for validation markers (#31)

jfaltermeier:
* introduce dedicated GIssueMarker which enabled the SDecoration default
features including hovering
* introduce GlspHoverMouseListener which is able to identify hovering
over GIssueMarkers and displays a popup without a server roundtrip
* introduce glspHoverModule which binds the GlspHoverMouseListener for
the MouseListener type and export it
* add additional css class on parent element when adding marker
* clear this again when marker is cleared

Examples:
https://github.com/eclipse-glsp/glsp-examples/pull/33 (+ server fix for examples https://github.com/eclipse-glsp/glsp-server/pull/43)

### DIFF
--- a/src/features/validation/validate.ts
+++ b/src/features/validation/validate.ts
@@ -30,6 +30,7 @@ import {
 
 import { GLSP_TYPES } from "../../types";
 import { Marker, MarkerKind } from "../../utils/marker";
+import { addCssClasses, removeCssClasses } from "../../utils/smodel-util";
 import { getSeverity } from "../hover/hover";
 import { IFeedbackActionDispatcher, IFeedbackEmitter } from "../tool-feedback/feedback-action-dispatcher";
 import { FeedbackCommand } from "../tool-feedback/model";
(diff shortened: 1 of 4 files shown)


src/features/validation/validate.ts
```diff
@@ -168,6 +171,24 @@ export class ApplyMarkersCommand extends FeedbackCommand {
     }
 }
 
+function addCSSClassToIssueParent(modelElement: SParentElement, issueMarker: SIssueMarker) {
```
tortmayr: You could reuse the utility mehtods for adding/removing cssClasses for this: https://github.com/eclipse-glsp/glsp-client/blob/db93ed9d9e6fa8926d662c45081edcc233d4de98/src/utils/smodel-util.ts#L61

src/features/validation/validate.ts
```diff
@@ -168,6 +171,24 @@ export class ApplyMarkersCommand extends FeedbackCommand {
...
+    if (modelElement.cssClasses === undefined) {
+        modelElement.cssClasses = [getSeverity(issueMarker)];
+    } else {
+        modelElement.cssClasses.push(getSeverity(issueMarker));
+    }
+}
+
+function removeCSSClassFromIssueParent(modelElement: SParentElement, issueMarker: SIssueMarker) {
```
tortmayr: Please use the utility method from `src/utils/smodel-utils.ts` for this

src/features/validation/validate.ts
jfaltermeier: Thanks, addressed in a follow-up commit

src/features/validation/validate.ts
jfaltermeier: Thanks, addressed in a follow-up commit

PR #38, opened 2020-02-17
Fix failing copy for SVG selection (glsp#22)

planger:
https://github.com/eclipse-glsp/glsp/issues/22

### DIFF
--- a/src/features/copy-paste/copy-paste-context-menu.ts
+++ b/src/features/copy-paste/copy-paste-context-menu.ts
@@ -1,0 +1,104 @@
+/********************************************************************************
+ * Copyright (c) 2020 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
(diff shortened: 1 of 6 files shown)


src/features/copy-paste/copy-paste-context-menu.ts
```diff
@@ -0,0 +1,104 @@
...
+    handle(action: Action) {
+        switch (action.kind) {
+            case InvokeCopyAction.KIND:
+                document.execCommand('copy');
+                break;
+            case InvokePasteAction.KIND:
+                // in a browser without additional permission we can't invoke the paste command
+                // the user needs to invoke it from the browser, so notify the user about it
+                this.dispatcher.dispatch(
+                    <GLSPServerStatusAction>{
```
martin-fleck-at: This shows a warning directly in the GLSP editor which is VERY subtle, I almost did not notice it. In Theia they open a message box. Not sure which way we should go with that but I just wanted to point that out ;-)

planger: Yes, thanks for raising that! This should actually be addressed with https://github.com/eclipse-glsp/glsp-theia-integration/pull/24

PR #51, opened 2020-04-01
Add auto-complete widget and generic edit validation support

planger:
The idea of the generic edit validation is to provide a generic mechanism for validating inputs while editing. In client projects we often introduce custom actions and action handlers for that, which all look the same but require different action types so that -- depending on context -- we can register different validators on the server. This change introduces one generic edit validation mechanism with a context ID that is used to switch between registered validators on the server. This is similar to how we do it with context action providers.

This change also harmonizes the existing label edit validation to use the same validation request and response actions as the generic one. The label edit validation is just a specific context (similar to command palette or tool palette for context actions).

https://github.com/eclipse-glsp/glsp/issues/56


src/base/auto-complete/auto-complete-actions.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+import { LabeledAction } from "sprotty/lib";
+
+export interface SetAutoCompleteValueAction extends LabeledAction {
+    readonly text: string;
+}
+
+export namespace SetAutoCompleteValueAction {
```
tortmayr, line 22: For consistency this should probably be reworked to an "isSetAutocompleteValueAction" function. In general i prefer the `Namespace.is()` approach but sprotty uses the isXXXAction() guard functions so we should do the same.

src/base/auto-complete/auto-complete-widget.ts
```diff
@@ -0,0 +1,304 @@
...
+import { IValidationDecorator } from "./validation-decorator";
+
+export interface AutoCompleteSettings {
+    readonly noSuggestionsMessage: string;
+    readonly suggestionsClass: string;
+    readonly debounceWaitMs: number;
+    readonly showOnFocus: boolean;
+}
+
+export abstract class InputValidator {
```
tortmayr, line 32: Is there a particular reason why you use abstract classes instead of interfaces here?

PR #55, opened 2020-04-21
Consider potential snapping in resize handles of change bounds tool

martin-fleck-at:
- Fixes eclipse-glsp/glsp/issues/23
resize-handle-snapping

Please note that with snapping the mouse and the handles do not necessarily align anymore. For example, a snapping to a grid of size 10 with rounding values normally leads to a snap of the resize handle on a move change > 5, therefore the handle moves 'faster' than the mouse which has only moved 5 pixel. 

EDIT: After the great suggestion from Tobi, I now hide the mouse cursor and resizing feels much smoother! It is a bit surprising when the mouse cursor reappears but that is not really a big issue. Additionally, I now also highlight the currently active 'dragging' cursor:

resize-handle-snapping-nocursor

### DIFF
--- a/css/glsp-sprotty.css
+++ b/css/glsp-sprotty.css
@@ -60,6 +60,10 @@
     fill: #66a;
 }
 
+.sprotty-resize-handle.active {
+    fill: var(--theia-menu-selectionBackground);
+}
+
 .sprotty-resize-handle.mouseover {
     stroke: #112;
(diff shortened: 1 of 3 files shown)


martin-fleck-at:
@tortmayr Great suggestion, I adapted the code accordingly and it feels much better, see also updated GIF in PR description!

PR #59, opened 2020-05-05
Use model source provider in editor context

planger:
This is to avoid an early dependency on the model source in case the editor context is used before the model source is bound.


planger:
For the reference, this is the error that we get in a proprietary project and that we would like to avoid with this change:
```
planner.js:62 Uncaught (in promise) Error: No matching bindings found for serviceIdentifier: Symbol(ModelSource)
    at _validateActiveBindingCount (planner.js:62)
    at _getActiveBindings (planner.js:48)
    at _createSubRequests (planner.js:91)
    at planner.js:115
    at Array.forEach (<anonymous>)
    at planner.js:114
    at Array.forEach (<anonymous>)
```

PR #75, opened 2020-08-15
#98 Remove theia-related css keys from @eclipse-glsp/client

tortmayr:
The css stylesheets in @eclispe-glsp/client should be application/framework independent.
If needed, they can be overwritten by css files that are defined in the corresponding framework integration package (e.g. theia-integration)
Fixes https://github.com/eclipse-glsp/glsp/issues/98

### DIFF
--- a/css/decoration.css
+++ b/css/decoration.css
@@ -1,11 +1,27 @@
+/********************************************************************************
+ * Copyright (c) 2020 EclipseSource and others.
+ *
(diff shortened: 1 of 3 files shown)


martin-fleck-at (approved):
Changes look good to me, there is just one thing missing from the migration to https://github.com/eclipse-glsp/glsp-theia-integration/pull/43/files but it doesn't affect this change, thank you Tobi!

PR #78, opened 2020-08-24
FeedbackCommands should rely on elementId rather than element reference

CamilleLetavernier:
- Fix ModifyCssFeedbackCommand to use elementId rather than element reference

### DIFF
--- a/src/features/tool-feedback/css-feedback.ts
+++ b/src/features/tool-feedback/css-feedback.ts
@@ -21,7 +21,12 @@ import { FeedbackCommand } from "./model";
 
 export class ModifyCSSFeedbackAction implements Action {
     kind = ModifyCssFeedbackCommand.KIND;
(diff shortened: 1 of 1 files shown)


tortmayr (changes requested):
Looks good to me.
However, I think it's better if we don't use direct element references in the action as well.
Otherwise subclassed/customized commands might use the element reference directly

CamilleLetavernier:
Should we keep the element references in the constructor (And just assign that a string[] elementIds field?), or should we completely remove the SModelElements from the Action's constructor and use directly ids?

The first option would have less impact on existing clients, but is maybe a little bit less consistent

tortmayr:
I'm fine with keeping the references in the constructor for now. I think we can life we the litte inconsistency and if needed we can easily fix this in a follow-up PR:

tortmayr (approved):
LGTM :+1:

PR #79, opened 2020-08-24
#94 #104 #96  Clean up communication protocol

tortmayr:
- Clean up communication and move implementation into glsp client
- Define a "clean" `GLSPClient` interface that is independent from the underlying communication protocol
- Provide a base implementation for a jsonrpc-based `GLSPClient`
- Update dependencies to sprotty 0.9.0
- Align dependency versions with Theia versions
- Add new DisposeClientAction to notify the server if a specific diagram client/widget can be disposed (e.g. on editor tab close)

Part of:
- eclipse-glsp/glsp/issues/104
- eclipse-glsp/glsp/issues/94
- eclipse-glsp/glsp/issues/96

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,14 +28,15 @@
   },
   "dependencies": {
     "autocompleter": "5.1.0",
-    "sprotty": "next",
+    "sprotty": "0.9.0",
     "uuid": "7.0.3",
-    "vscode-ws-jsonrpc": "^0.0.2-1"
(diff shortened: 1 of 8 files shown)


planger (approved):
Great work, thanks a lot!
I've appended a few minor changes in a separate commit for your review. Other than that, this looks great!
Just a general remark regarding copyrights, please keep the original where applicable but just update the year (e.g. original 2019 --> now 2019 - 2020). The copyright owner should remain the same.

PR #80, opened 2020-08-25
Use new GLSPClient protocl in standalone context

tortmayr:
- Refactor Jsonrpc-based protocol implementation into sub directory
- Replace GLSPWebsocketDiagramServer with GLSPDiagramServer. The GlspDiagramServer uses the newly introduced GLSPClient protocol.
- Restructure features/select and features/validate to get rid of circular dependency warnings


planger (approved):
LGTM, thanks!

PR #85, opened 2020-09-07
Migrate getPort function into @eclipse-glsp/protocol

tortmayr:
This enables reuse outside of Theia e.g. in VS Code


martin-fleck-at (approved):
Looks good to me, thank you Tobias!

PR #86, opened 2020-09-23
#110 Improve Palette-tool disabling

CamilleLetavernier:
The Palette Tool is now focused when it's activated, and it can handle the Escape key-press (Thus disabling itself).


martin-fleck-at (approved):
Change looks good to me and works great, thank you Camille! Finally people are allowed to make wrong clicks again :P The only thing I noticed, as already mentioned by you, is that the first Escape is not recognized at all. This is not specific to the palette or your change but very curious indeed.

PR #87, opened 2020-09-24
#123 Centralize configuration in the server

CamilleLetavernier:
Use the new ConfigureServerHandlersAction capability, to dynamically configure server-handled actions, rather than listing them explicitly. This action contains the list of actionKinds that the server can handled, and is returned by the server during the InitializeClientSessionAction phase.

Some actions still need to be registered explicitly by the client, as they happen during the InitializeClientSession phase. If we want to avoid this, we'd have to wait for the ConfigureServerHandlersAction before requesting a model; but I don't believe this is necessary (As we now have to simply deal with a fixed number of initialization actions, which are not likely to evolve much anyway. All custom or feature-specific actions are dynamically registered)

Depends on https://github.com/eclipse-glsp/glsp-server/pull/80 for the server-side changes

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -4,12 +4,10 @@
   "version": "2.0.0",
   "tasks": [
     {
-      "label": "Build all",
+      "label": "[Client] Build all",
+      "detail": "Build all glsp-client packages & examples",
       "type": "shell",
-      "group": {
-        "kind": "build",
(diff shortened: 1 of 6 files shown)


planger (approved):
Thank you very much! This change greatly avoids unnecessary customizations of client code that is only caused by adding custom actions on the server!

PR #72, opened 2020-07-17
#87 Add readonly-support for GLSP diagrams

tortmayr:
- Introduce SetEditModeAction, make EditorContextService editModeAware & add editModeChangeListener support
- Refactor Tool API to determine tools with edit-functionality.
- Remove code-duplication by extracting a common tool base class
- Refactor Tools to avoid constructor injection
- Convert CommandPalette & EditLabel feature into tools
- Disable commandPalette/contextMenu items with edit function if in readonly mode
- Improve toolPalette css to avoid unintended text-selection

part of  eclipse-glsp/glsp/issues/87

PR #88, opened 2020-09-28
#66, #72: Refactor code for easier customization

martin-fleck-at:
- Use property injection over constructor injection
- Avoid readonly properties in classes that should be customizable

Fixes https://github.com/eclipse-glsp/glsp/issues/66
Fixes https://github.com/eclipse-glsp/glsp/issues/72

### DIFF
--- a/packages/client/src/features/change-bounds/movement-restrictor.ts
+++ b/packages/client/src/features/change-bounds/movement-restrictor.ts
@@ -45,7 +45,7 @@ export function removeMovementRestrictionFeedback(element: SModelElement, moveme
 
 @injectable()
 export class NoOverlapMovmentRestrictor implements IMovementRestrictor {
-    public cssClasses = ["movement-not-allowed"];
+    cssClasses = ["movement-not-allowed"];
 
     validate(newLocation: Point, element: SModelElement): boolean {
         if (!isBoundsAwareMoveable(element)) {
--- a/packages/client/src/features/change-bounds/movement-restrictor.ts
+++ b/packages/client/src/features/change-bounds/movement-restrictor.ts
@@ -45,6 +45,8 @@ export function removeMovementRestrictionFeedback(element: SModelElement, moveme
 
 @injectable()
 export class NoOverlapMovmentRestrictor implements IMovementRestrictor {
+    public cssClasses = ["movement-not-allowed"];
+
     validate(newLocation: Point, element: SModelElement): boolean {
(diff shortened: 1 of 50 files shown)


packages/client/src/features/change-bounds/movement-restrictor.ts
```diff
@@ -45,6 +45,8 @@ export function removeMovementRestrictionFeedback(element: SModelElement, moveme
 
 @injectable()
 export class NoOverlapMovmentRestrictor implements IMovementRestrictor {
+    public cssClasses = ["movement-not-allowed"];
```
tortmayr, line 48: [Very,very minor]: Why did you opt for the explicit `public` modifier her? AFAIK we omit `public` in most other cases and should probably keep it consistent.

packages/client/src/features/validation/validate.ts
```diff
@@ -94,7 +93,7 @@ export function isSetMarkersAction(action: Action): action is SetMarkersAction {
 @injectable()
 export abstract class ExternalMarkerManager {
 
-    languageLabel: string;
+    public languageLabel: string;
```
tortmayr, line 96: Implicit vs. explicit public (consistency)

tortmayr (approved):
Thanks a lot. Code looks good to me. I also appreciate the additional effort to identify unused arguments and prefixing them with `_`.  :wink: 

I have raised one minor issue regarding implicit vs. explicit `public` but that does not prevent merging.
:+1:

packages/client/src/features/change-bounds/movement-restrictor.ts
martin-fleck-at: Good point! I believe I added it explicitly as I had the feeling that we do add the `public` modifier more often than not, e.g., in all action constructors we use the public keyword. I am fine with both but it should be consistent. It is probably my Java background that tilts me a little bit more towards an explicit 'public' but I am also fine with omitting it. What do you think?

tortmayr: Constructors are a different story, constructor variables are private by default, here you need the explicit `public`.
I think we should opt for implicit public were applicable. We are heavily reusing code from Theia and sprotty and both use implicit public as well. So I guess, we should do the same.

martin-fleck-at: Good point! I rebased the branch and removed the explicit modifiers in a separate commit. Thank you for the explanation!

PR #91, opened 2020-10-29
Handle model source change notification

planger:
https://github.com/eclipse-glsp/glsp/issues/134


martin-fleck-at (approved):
Smooth PR, all looks good to me and works well with the server change, thank you Philip!

PR #95, opened 2020-12-01
#153: Improve external navigation support through dedicated action

martin-fleck-at:
- Introduce dedicated 'NavigateToExternalTarget' action
- Replace external target service with action handler

Part of eclipse-glsp/glsp/issues/153


planger (approved):
Great, this makes the external navigation way more flexible! Thanks Martin!

PR #97, opened 2020-12-18
[136] Readme for Workflow example and integrations

CamilleLetavernier:


README.md
```diff
@@ -2,6 +2,35 @@
...
+
+```
+yarn install
+```
+
+And the GLSP Server (Which includes the Workflow Diagram Server example). See [`glsp-server`](https://github.com/eclipse-glsp/glsp-server#building) for instructions.
+
+Once both the Server and the Client packages are built, you can start both:
+
+- Server: in `examples/org.eclipse.glsp.example.workflow`, start `org.eclipse.glsp.example.workflow.launch.WorkflowServerLauncher.java` with these arguments: `--websocket --port=8081` (Or use the included `Start_Workflow_Example_Server_(Websocket).launch` Launch Configuration from Eclipse).
```
planger, line 25: Maybe it is worth mentioning that the server can also be simply started from command line after building as documented here: https://github.com/eclipse-glsp/glsp#building-and-running-the-glsp-backend
This way, interested users wouldn't have to set up an IDE?

README.md
```diff
@@ -2,6 +2,35 @@
...
+- Client: Open the `examples/workflow-standalone/app/diagram.html` file in your favorite browser
+
+### Where to find the sources?
+
+In addition to this repository, the related source code can be found here:
```
planger, line 30:
```suggestion
In addition to this repository, the source code of the example diagram server can be found here: https://github.com/eclipse-glsp/glsp-server/tree/master/examples/org.eclipse.glsp.example.workflow
```

README.md
```diff
@@ -2,6 +2,35 @@
...
+
+-   https://github.com/eclipse-glsp/glsp-server
```
planger, line 32: I'd prefer just mentioning the link in the paragraph above instead of having a single-bullet bullet list.

planger (commented):
Thanks, Camille! Looks good to me!
Just two suggestions below. What do you think?

README.md
CamilleLetavernier: There are too many readmes spread all around :) I wasn't aware of this part of the doc

README.md
CamilleLetavernier: I followed the structure for all other READMEs (The glsp-client is the only one with a single related repo); but removing the list is also fine

planger (approved):

PR #102, opened 2021-02-05
Add focus tracking and remove hover on lost focus

planger:
Re https://github.com/eclipse-glsp/glsp/issues/184


martin-fleck-at (approved):
Change looks good and works very well, thank you Philip!

PR #103, opened 2021-02-11
Generalize input for ModifyCSSFeedbackAction

tortmayr:
Adapt `ModifyCSSFeedbackAction` to support both a string[] with the smodel ids or a SModelElement[] as input.

### DIFF
--- a/packages/client/src/features/tool-feedback/css-feedback.ts
+++ b/packages/client/src/features/tool-feedback/css-feedback.ts
@@ -16,6 +16,7 @@
 import { inject, injectable } from "inversify";
 import { Action, CommandExecutionContext, SModelElement, SModelRoot, TYPES } from "sprotty";
 
+import { isStringArray } from "../../utils/array-utils";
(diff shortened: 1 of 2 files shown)


packages/client/src/features/tool-feedback/css-feedback.ts
```diff
@@ -23,16 +23,23 @@ export class ModifyCSSFeedbackAction implements Action {
...
         public kind = ModifyCssFeedbackCommand.KIND) {
-        if (elements) {
-            this.elementIds = elements.map(elt => elt.id);
+        if (input) {
+            this.elementIds = isStringArray(input) ? input : input.map(element => element.id);
         }
     }
 }
 
+function isStringArray(value: object): value is string[] {
```
martin-fleck-at, line 36: I think this method has high re-usability, do you think we can move it in a utility class somewhere?

tortmayr: Done

martin-fleck-at (approved):
Change looks good to me, thank you Tobias!

PR #104, opened 2021-02-23
Make auto complete widget more flexible

planger:
Allow to change auto complete settings, suggestion provider, and
suggestion submit handler after instantiation. This is often used
in the onBeforeShow of UI extensions using the auto complete widget.
In the onBeforeShow we have more context and may want to change the
settings without having to reinstantiate the entire widget.


martin-fleck-at (approved):
Looks good to me, thank you Philip!

PR #105, opened 2021-02-23
Upgrade Sprotty to fix inversify version mismatch

planger:
Sprotty had a hard dependency on 5.0.1 which forced glsp-client to
also depend hard on 5.0.1. Theia however has ^5.0.1 which resolves
to 5.0.5, so we had a version mismatch.
We fixed this dependency in Sprotty and now we can upgrade the
Sprotty dependency in glsp-client to resolve the version mismatch
between glsp-client and glsp-theia-integration.


tortmayr (approved):
LGTM!

PR #106, opened 2021-02-23
Add SArgumentable interface denoting elements that contain args

planger:
https://github.com/eclipse-glsp/glsp/issues/194

### DIFF
--- a/packages/client/src/base/args.ts
+++ b/packages/client/src/base/args.ts
@@ -21,6 +21,6 @@ export interface SArgumentable {
     args: Args;
 }
 
-export function hasArguments(element: SModelElement | undefined): element is SModelElement & SArgumentable {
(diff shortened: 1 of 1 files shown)


packages/client/src/base/args.ts
```diff
@@ -13,4 +13,14 @@
...
  ********************************************************************************/
+import { SModelElement } from "sprotty";
+
 export type Args = { [key: string]: string | number | boolean };
+
+export interface SArgumentable {
+    args: Args;
+}
+
+export function hasArguments(element: SModelElement | undefined): element is SModelElement & SArgumentable {
```
tortmayr, line 24: I would opt for the concise notation of optional arguments here:
```suggestion
export function hasArguments(element?: SModelElement): element is SModelElement & SArgumentable {
```

tortmayr (approved):

PR #108, opened 2021-03-18
Add marquee tool

sgraband:
Add marquee tool to select multiple elements at once. Selecting with the marquee tool is done by clicking and dragging the mouse while the shift key is pressed.

MarqueeFinished

### DIFF
--- a/examples/workflow-glsp/src/model.ts
+++ b/examples/workflow-glsp/src/model.ts
@@ -27,6 +27,7 @@ import {
     moveFeature,
     Nameable,
     nameFeature,
+    Point,
     popupFeature,
     RectangularNode,
     SEdge,
@@ -56,6 +57,13 @@ export class TaskNode extends RectangularNode implements Nameable, WithEditableL
         }
         return undefined;
     }
+}
+
+export class MarqueeNode extends RectangularNode {
+    static readonly DEFAULT_FEATURES = [connectableFeature, deletableFeature, selectFeature, boundsFeature,
+        moveFeature, layoutContainerFeature, fadeFeature, hoverFeedbackFeature, popupFeature, nameFeature, withEditLabelFeature];
+    startPoint: Point;
+    endPoint: Point;
 }
 
 export function isTaskNode(element: SModelElement): element is TaskNode {
--- a/examples/workflow-glsp/src/model.ts
+++ b/examples/workflow-glsp/src/model.ts
@@ -27,7 +27,6 @@ import {
     moveFeature,
     Nameable,
     nameFeature,
-    Point,
     popupFeature,
     RectangularNode,
     SEdge,
@@ -57,13 +56,6 @@ export class TaskNode extends RectangularNode implements Nameable, WithEditableL
         }
         return undefined;
     }
-}
-
-export class MarqueeNode extends RectangularNode {
-    static readonly DEFAULT_FEATURES = [connectableFeature, deletableFeature, selectFeature, boundsFeature,
-        moveFeature, layoutContainerFeature, fadeFeature, hoverFeedbackFeature, popupFeature, nameFeature, withEditLabelFeature];
-    startPoint: Point;
-    endPoint: Point;
 }
 
 export function isTaskNode(element: SModelElement): element is TaskNode {
--- a/examples/workflow-glsp/src/workflow-views.tsx
+++ b/examples/workflow-glsp/src/workflow-views.tsx
@@ -28,7 +28,7 @@ import { injectable } from "inversify";
 import * as snabbdom from "snabbdom-jsx";
 import { VNode } from "snabbdom/vnode";
 
-import { ActivityNode, Icon, TaskNode, WeightedEdge } from "./model";
+import { ActivityNode, Icon, MarqueeNode, TaskNode, WeightedEdge } from "./model";
 
 const JSX = { createElement: snabbdom.svg };
 
@@ -60,6 +60,21 @@ export class ForkOrJoinNodeView extends RectangularNodeView {
             <rect class-sprotty-node={true} class-forkOrJoin={true}
                 class-mouseover={node.hoverFeedback} class-selected={node.selected}
                 width={10} height={Math.max(50, node.bounds.height)}></rect>
+        </g>;
+        return graph;
+    }
+}
+
+@injectable()
+export class MarqueeView extends RectangularNodeView {
+    render(node: MarqueeNode, context: RenderingContext): VNode {
+        const graph = <g>
+            <rect class-sprotty-node={true} class-marquee={true}
+                x={((node.startPoint.x - node.endPoint.x) <= 0) ? node.startPoint.x : node.endPoint.x}
+                y={((node.startPoint.y - node.endPoint.y) <= 0) ? node.startPoint.y : node.endPoint.y}
+                rx={0} ry={0}
+                width={Math.abs(node.startPoint.x - node.endPoint.x)}
+                height={Math.abs(node.startPoint.y - node.endPoint.y)}></rect>
         </g>;
         return graph;
     }
--- a/examples/workflow-glsp/src/workflow-views.tsx
+++ b/examples/workflow-glsp/src/workflow-views.tsx
@@ -28,7 +28,7 @@ import { injectable } from "inversify";
 import * as snabbdom from "snabbdom-jsx";
 import { VNode } from "snabbdom/vnode";
 
-import { ActivityNode, Icon, MarqueeNode, TaskNode, WeightedEdge } from "./model";
+import { ActivityNode, Icon, TaskNode, WeightedEdge } from "./model";
 
 const JSX = { createElement: snabbdom.svg };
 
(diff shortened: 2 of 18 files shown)


planger:
Just one comment from the top of my head, I think it'd be cool to also add a tool to the palette to enable the marquee tool directly. Can be done in a follow-up PR though.

examples/workflow-glsp/src/model.ts
```diff
@@ -58,6 +59,13 @@ export class TaskNode extends RectangularNode implements Nameable, WithEditableL
     }
 }
 
+export class MarqueeNode extends RectangularNode {
+    static readonly DEFAULT_FEATURES = [connectableFeature, deletableFeature, selectFeature, boundsFeature,
```
tortmayr, line 63: I think we should remove all unneeded features here to avoid potential unintended side effects.
IMO it should be enough to only activate the `boundsFeature`

examples/workflow-glsp/src/workflow-views.tsx
```diff
@@ -65,6 +65,21 @@ export class ForkOrJoinNodeView extends RectangularNodeView {
     }
 }
 
+@injectable()
```
tortmayr, line 68: The view definition should be part of `@eclipse-glsp/client` to enable generic reuse.

packages/client/src/features/tools/di.config.ts
```diff
@@ -23,15 +23,19 @@ import { DelKeyDeleteTool, MouseDeleteTool } from "./delete-tool";
...
 import { NodeCreationTool } from "./node-creation-tool";
+import { MarqueeMouseTool } from "./marquee-mouse-tool";
+import { MarqueeTool } from "./marquee-tool";
 
 const toolsModule = new ContainerModule((bind, _unbind, isBound) => {
     // Register default tools
     bind(GLSP_TYPES.IDefaultTool).to(ChangeBoundsTool);
     bind(GLSP_TYPES.IDefaultTool).to(EdgeEditTool);
     bind(GLSP_TYPES.IDefaultTool).to(DelKeyDeleteTool);
+    bind(GLSP_TYPES.IDefaultTool).to(MarqueeTool);
```
tortmayr, line 34: It probably makes sense to introduce a dedicated 'marqueeToolModule'
This module than binds the corresponding tools, and configures the modelelement/view.

packages/client/src/features/tools/marquee-mouse-tool.ts
```diff
@@ -0,0 +1,208 @@
...
+import { DrawMarqueeAction, RemoveMarqueeAction } from "../tool-feedback/selection-tool-feedback";
+import { BaseGLSPTool } from "./base-glsp-tool";
+import { getAbsolutePosition, toAbsoluteBounds } from "../../utils/viewpoint-util";
+import { DOMHelper } from "sprotty/lib/base/views/dom-helper";
+
+@injectable()
+export class MarqueeMouseTool extends BaseGLSPTool {
+    static ID = "glsp.marquee-mouse-tool";
+
+    @inject(GLSP_TYPES.MouseTool) protected mouseTool: IMouseTool;
```
tortmayr, line 32: The mouseTool,keyTool and feedbackActionDispatcher are already injected in the super class ('BaseGLSPTool'). No need to inject them here again

packages/client/src/features/tools/marquee-mouse-tool.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    get id(): string {
+        return MarqueeMouseTool.ID;
+    }
+
+    enable(): void {
+        this.marqueeMouseListener = new MarqueeMouseListener(this.domHelper);
+        this.mouseTool.register(this.marqueeMouseListener);
+        this.keyTool.register(this.shiftKeyListener);
+        this.feedbackDispatcher.registerFeedback(this, [cursorFeedbackAction(CursorCSS.MARQUEE)]);
```
tortmayr, line 48: You can use the `dispatchFeedback()` method from the super class for this

packages/client/src/features/tools/marquee-mouse-tool.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    disable(): void {
+        this.mouseTool.deregister(this.marqueeMouseListener);
+        this.keyTool.deregister(this.shiftKeyListener);
+        this.feedbackDispatcher.registerFeedback(this, [cursorFeedbackAction()]);
```
tortmayr, line 54: You can use the `deregisterFeedback()' method from the superclass for this

packages/client/src/features/tools/marquee-tool.ts
```diff
@@ -0,0 +1,51 @@
...
+import { inject, injectable } from "inversify";
+import { Action, EnableToolsAction, KeyListener, KeyTool, SModelElement } from "sprotty";
+import { BaseGLSPTool } from "./base-glsp-tool";
+import { MarqueeMouseTool } from "./marquee-mouse-tool";
+
+@injectable()
+export class MarqueeTool extends BaseGLSPTool {
+    static ID = "glsp.marquee-tool";
+
+    protected selectionKeyListener: MarqueeKeyListener = new MarqueeKeyListener();
```
tortmayr, line 25: Should probably be named 'marqueeKeyListener' instead of 'selectionKeyListener'

tortmayr (changes requested):
Thanks! In general the tool works well and is a great contribution to glsp.
I have a couple of inline comments and found one general issue:

**The current implementation is specific to the workflow-example and cannot be reused generically.** 
We should move the model configuration and the marquee view definition into the glsp-client package. It is probably a good idea to create a dedicated `marqueeToolsModule` that handles  both the tool registration/binding and the configuration of the marqueeView.

packages/client/src/features/tools/di.config.ts
sgraband: So that the marquee tool is no longer handled by 'tools' feature and is completly handled in a newly created 'marquee-tool' feature?
The marqueeToolModule would then look like this:
`const glspMarqueeToolModule = new ContainerModule((bind, unbind, isBound, rebind) => {
    configureModelElement(context, 'marquee', MarqueeNode, MarqueeView);
    bind(GLSP_TYPES.IDefaultTool).to(MarqueeTool);
    bind(GLSP_TYPES.ITool).to(MarqueeMouseTool);
});`

tortmayr: Correct, however on second thought it may be better to just create a utility function `configureMarqueeTool` which is then called inside of the tools module
`const toolsModule = new ContainerModule((bind, _unbind, isBound) => {
...
 const context = { bind, isBound};
configureMarqueeTool(context);
}

export function configureMarqueeTool((context: { bind: interfaces.Bind, isBound: interfaces.IsBound }){

    configureModelElement(context, 'marquee', MarqueeNode, MarqueeView);
    bind(GLSP_TYPES.IDefaultTool).to(MarqueeTool);
    bind(GLSP_TYPES.ITool).to(MarqueeMouseTool);
'

sgraband: Would you then expect the `MarqueeNode`, `MarqueeView`, `MarqueeMouseTool` and `MarqueeTool` to be in the tools feature or in the separate marquee-tool, even though they will be bound in the tools feature?

tortmayr: I would omit the separate marque-tool for now and keep  MarqueeNode, MarqueeView, MarqueeMouseTool and MarqueeTool in the tools feature but extract the actual binding process into a utility function. This way it can be easily reused and/or removed in subclasses.

packages/client/src/features/tools/di.config.ts
```diff
@@ -13,36 +13,47 @@
...
     // Register  tools
     bind(GLSP_TYPES.ITool).to(MouseDeleteTool);
-    bind(GLSP_TYPES.ITool).to(MarqueeMouseTool);
     bind(NodeCreationTool).toSelf().inSingletonScope();
     bind(EdgeCreationTool).toSelf().inSingletonScope();
     bind(GLSP_TYPES.ITool).toService(EdgeCreationTool);
     bind(GLSP_TYPES.ITool).toService(NodeCreationTool);
 
+    configureMarqueeTool({ bind, isBound });
+    configureActionHandler({ bind, isBound }, EnableToolsAction.KIND, GLSPScrollMouseListener);
```
tortmayr, line 47: Should be moved into the `glspViewportModule`

packages/client/src/features/tools/di.config.ts
```diff
@@ -13,36 +13,47 @@
...
+    configureActionHandler({ bind, isBound }, EnableDefaultToolsAction.KIND, GLSPScrollMouseListener);
```
tortmayr, line 48: Should be moved into the `glspViewportModule`

tortmayr (changes requested):
Thanks a lot. Change looks good to me now, I only have one minor suggestion

tortmayr (approved):
LGTM :+1:

PR #113, opened 2021-04-16
#180: Provide default model and view registrations and add views

martin-fleck-at:
- Registrations
-- Provide default types as strings
-- Register elements for default types
-- Register views for default types
-- Ensure element and view registrations can be overwritten

- Views
-- Add rounded corner view that reads argument map for corner radius
-- Add edge view that supports custom padding for easier mouse handling
-- Move issue marker view to shared views folder

- Use custom edge and rounded corner view in example for new elements

Fixes https://github.com/eclipse-glsp/glsp/issues/180

### DIFF
--- a/examples/workflow-glsp/src/di.config.ts
+++ b/examples/workflow-glsp/src/di.config.ts
@@ -20,15 +20,16 @@ import 'sprotty/css/edit-label.css';
 import {
     boundsModule,
     buttonModule,
+    configureDefaultModelElement,
     configureModelElement,
     ConsoleLogger,
     defaultGLSPModule,
     defaultModule,
+    DefaultTypes,
     DeleteElementContextMenuItemProvider,
     DiamondNodeView,
     edgeLayoutModule,
     editLabelFeature,
-    ExpandButtonView,
     expandModule,
     exportModule,
     fadeModule,
@@ -37,15 +38,12 @@ import {
     glspContextMenuModule,
     glspDecorationModule,
     glspEditLabelModule,
-    GLSPGraph,
     glspHoverModule,
     glspMouseToolModule,
     glspSelectModule,
     glspServerCopyPasteModule,
     glspViewportModule,
     GridSnapper,
-    HtmlRoot,
-    HtmlRootView,
     labelEditUiModule,
     layoutCommandsModule,
     LogLevel,
@@ -58,19 +56,13 @@ import {
     openModule,
     overrideViewerOptions,
     paletteModule,
-    PreRenderedElement,
-    PreRenderedView,
     RevealNamedElementActionProvider,
     routingModule,
-    SButton,
     SCompartment,
     SCompartmentView,
     SEdge,
-    SGraphView,
     SLabel,
     SLabelView,
-    SRoutingHandle,
-    SRoutingHandleView,
     toolFeedbackModule,
     toolsModule,
     TYPES,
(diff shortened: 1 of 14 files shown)


examples/workflow-glsp/src/workflow-views.tsx
```diff
@@ -34,24 +38,32 @@ import { ActivityNode, Icon, TaskNode, WeightedEdge } from './model';
...
-        const rcr = this.getRoundedCornerRadius(node);
+export class TaskNodeView extends RoundedCornerNodeView {
+    protected renderWithoutRadius(node: Readonly<SShapeElement & Hoverable & Selectable>, context: RenderingContext): VNode {
+        const task = node as TaskNode;
+        const rcr = this.getRoundedCornerRadius(task);
         const graph = <g>
-            <rect class-sprotty-node={true} class-task={true}
-                class-automated={node.taskType === 'automated'}
-                class-manual={node.taskType === 'manual'}
-                class-mouseover={node.hoverFeedback} class-selected={node.selected}
```
tortmayr, line 44: By removing `class-selected={node.selected}` we lose the selection feedback on task nodes

examples/workflow-glsp/src/workflow-views.tsx
```diff
@@ -67,36 +79,28 @@ export class ForkOrJoinNodeView extends RectangularNodeView {
...
-            class-weighted={true}
-            class-low={edge.probability === 'low'}
-            class-medium={edge.probability === 'medium'}
-            class-high={edge.probability === 'high'}
-            class-mouseover={edge.hoverFeedback}>
-            {this.renderLine(edge, route, context)}
-            {this.renderAdditionals(edge, route, context)}
-            {context.renderChildren(edge, { route })}
-        </g>;
+    protected addtionalClasses(edge: Readonly<SEdge>, _context: RenderingContext): Classes {
```
tortmayr, line 96: Should be reworked with eclipse-glsp/glsp/issues/220 and applied with the  `GModelElement.cssClasses` property

examples/workflow-glsp/src/workflow-views.tsx
```diff
@@ -34,24 +38,32 @@ import { ActivityNode, Icon, TaskNode, WeightedEdge } from './model';
...
 
-    protected getRoundedCornerRadius(node: SShapeElement): number {
+    protected getRoundedCornerRadius(_node: TaskNode): number {
         return 5;
     }
+
+    protected additionalClasses(element: Readonly<SShapeElement & Hoverable & Selectable>, context: RenderingContext): Classes {
+        const node = element as TaskNode;
+        return {
+            'class-task': true,
```
tortmayr, line 62: Should be reworked with eclipse-glsp/glsp/issues/220 and applied with the GModelElement.cssClasses property

tortmayr (changes requested):
Looks good to me in general. However, with this change the selection feedback for tasknodes does not work anymore. Other than that I just have a few minor nitpicks

examples/workflow-glsp/src/workflow-views.tsx
martin-fleck-at: Good catch! In the newly created manual nodes it works because it is considered in the rounded edge view but in the existing views loaded from the model file we lose it since I "optimized" those classes away. I re-added them for the non-corner case ;-)

packages/client/src/views/base-view-module.ts
```diff
@@ -0,0 +1,98 @@
...
+} from 'sprotty';
+
+import { GLSPGraph } from '../lib/model';
+import { DefaultTypes } from './default-types';
+import { GEdgeView } from './glsp-edge-view';
+import { GIssueMarkerView } from './issue-marker-view';
+import { RoundedCornerNodeView } from './rounded-corner-view';
+
+interface ContainerContext {
+    bind: interfaces.Bind;
```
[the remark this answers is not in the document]
martin-fleck-at: The `configureModelElement` from Sprotty which we call only needs those two. Since this is not an exported interface anyway, I believe we can leave it as is, what do you think?

examples/workflow-glsp/src/di.config.ts
```diff
@@ -91,19 +83,15 @@ const workflowDiagramModule = new ContainerModule((bind, unbind, isBound, rebind
     bind(TYPES.ICommandPaletteActionProvider).to(RevealNamedElementActionProvider);
     bind(TYPES.IContextMenuItemProvider).to(DeleteElementContextMenuItemProvider);
     const context = { bind, unbind, isBound, rebind };
-    configureModelElement(context, 'graph', GLSPGraph, SGraphView);
+
+    configureDefaultModelElement(context);
```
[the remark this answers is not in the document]
martin-fleck-at: Good catch!

examples/workflow-glsp/src/workflow-views.tsx
```diff
@@ -34,24 +38,32 @@ import { ActivityNode, Icon, TaskNode, WeightedEdge } from './model';
 const JSX = { createElement: snabbdom.svg };
 
 @injectable()
```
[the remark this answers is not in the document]
martin-fleck-at: Great idea!

tortmayr (approved):
LGTM! Thanks :+1:

PR #115, opened 2021-05-03
#228: Introduce GlspSelectionDataService

ndoschek:
Fixes https://github.com/eclipse-glsp/glsp/issues/228


tortmayr (commented):
Changes look good to me. However, I'm wondering wether we should move this into the glsp-theia-integration. IMO the `GlspSelectionDataService` is not generally reusable, and is currently just used in the ´TheiaSelectionForwarder` so maybe the specification should also be placed there.

Also a No-op default binding doesn't seem to useful to me. Maybe it's better to use optional injection in `TheiaSelectionForwarder`.

ndoschek:
That's a good point, thanks Tobias! I will move it to `glsp-theia-integration` and close this PR in favor of https://github.com/eclipse-glsp/glsp-theia-integration/pull/69

PR #117, opened 2021-05-03
#227 Use concrete next version instead of tag

tortmayr:
Also add  update script to root package.json. The update script can be executed with `yarn upgrade:next`

Part of eclipse-glsp/glsp/issues/227


planger (approved):

PR #118, opened 2021-05-13
Revert #227

tortmayr:
Revert back to using next tags after identifying issues for glsp-dependent projects that occurred with eclipse-glsp/glsp/issues/227.
We still want to maintain a "yarn upgrade:next" script in all repositories.


martin-fleck-at (approved):
LGTM!

PR #120, opened 2021-05-18
Keep selection when using CTRL and marquee tool

sgraband:
Pressing CTRL now keeps the previous selection when using the marquee tool

Signed-off-by: Simon Graband <sgraband@eclipsesource.com>


planger (changes requested):
Thanks a lot! Works well! I'd just like to ask you to remove the console logs.

planger (approved):

PR #123, opened 2021-05-25
#224 Fix wrong feedback anchor computation for nested elements

tortmayr:
Fix feedback edge snapping for nested elements inside of a container. 
Compute the absolute anchor position (relative to the root graph) instead of the local anchor position (relative to the child element). This ensure that the edge snaps to the correct position.

This fixes the third subissue mentioned in eclipse-glsp/glsp/224

Also: Git ignore example server jar


planger (approved):

PR #127, opened 2021-07-07
#253 Align operation definitions for ChangeContainerOperation

ndoschek:
The definition of the optional member variable location in ChangeContainerOperation does not match on server and client side.

Please see also server change: https://github.com/eclipse-glsp/glsp-server/pull/115

Fixes https://github.com/eclipse-glsp/glsp/issues/253


tortmayr (approved):
LGTM! Thanks. :+1:

PR #130, opened 2021-08-10
#258 Remove name property from GLSPClient

tortmayr:
A client is already uniquely identified via its id, there is no need for an additional name property. In our current implementing projects the name property is always equal to the id and therefore redundant.

Part of eclipse-glsp/glsp/issues/258


planger (approved):

PR #132, opened 2021-08-12
#315 Refactor base protocol

tortmayr:
Refactor the base communication protocol to support initializing and disposing a client session for a given language. The method parameters are encapsulated in dedicated parameter objects to provide a stable API. Refactor the existing `initializeServer` method to return void instead of a boolean and use a generic args map for custom arguments instead of a dedicated object. In addition, the protocol version that is implemented by the client is now passed as part of the `InitializeParameters`

Refactor the standlone example to properly implement the new protocol.

Also:
- Remove no longer needs actions for initializing/disposing a client session.

Requires: https://github.com/eclipse-glsp/glsp-server/pull/123
Part of eclipse-glsp/glsp/issues/315


planger (approved):

tortmayr:
Thanks for the review Philip, I have made some additional changes to the protocol. Could you please re-review?

examples/workflow-standalone/src/main.ts
```diff
@@ -16,45 +16,67 @@
...
+    await client.initializeClientSession({ clientSessionId: diagramServer.clientId, diagramType });
+    actionDispatcher.dispatch(new RequestModelAction({
+        sourceUri: `file://${examplePath}`,
+        diagramType
+    }));
+    actionDispatcher.dispatch(new RequestTypeHintsAction(diagramType));
+    actionDispatcher.dispatch(new EnableToolPaletteAction());
+}
+
+async function configureServerActions(result: InitializeResult): Promise<void> {
```
planger, line 72: Shouldn't this function and maybe even the entire `initialize` function be generic enough to push it into the generic base framework?

planger (approved):

examples/workflow-standalone/src/main.ts
tortmayr: Good catch. I don't think the `initialize` function itself is generic enough but the `configureServerActions` definitly is. I moved it into the base framework and reused it both here and in the glsp-theia-integration

PR #134, opened 2021-08-20
#376 Element dragging without selection might not work with Firefox

ndoschek:
- Fix element movement without selection for Firefox by avoiding reset of `startDragPosition` on mouse entering (similar to other Firefox fixes in sprotty)

Fixes https://github.com/eclipse-glsp/glsp/issues/376


tortmayr (approved):
LGTM! Thanks

PR #150, opened 2021-11-01
Fix isSetEditValidationResultAction() function

planger:
This function wrongly checked for property `response`, while it actually should check for `repsonseId`.


tortmayr (approved):
Good catch. LGTM!

PR #152, opened 2021-11-05
#450 Improve Marquee tool (Performance & Full element marquee behavior)

ivy-lli:
closes https://github.com/eclipse-glsp/glsp/issues/450

- Refactor code
  - Move calculation code to new class for easy testing
  - Return direct expression instead in if else
- Improve performance
  - Cache possible elements at start, instead of collect them on each mouse move event
  - Calculate absolute positions only once
- Add possibility to change the select behavior to marque full element
- Fix run test launch configs
- Add tests for calculation code of marquee tool

New marquee behavior:
glsp-mark-entire-element

I did again some performance measures in the MarqueeMouseListener class (same as in the Issue discribed):
* touch behavior: avg. 0.22880000030994416 ms
* touch behavior with 6x cpu slowdown: avg. 0.649664429910231 ms
* entire behavior: avg. 0.1789156626236726 ms
* entire behavior with 6x cpu slowdown: avg. 0.5752136751245229 ms

So it seems that the refactoring did its work 😀

Now the behavior is tested and can be changed over DI, e.g by:
`container.bind(GLSP_TYPES.IMarqueeBehavior).toConstantValue({ entireEdge: true, entireElement: true });`

Please add feedback if you want to be something changed or have an other improvement 😉

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,9 +6,12 @@
       "request": "launch",
       "name": "Run current test",
       "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
-      "args": ["--opts", "${workspaceFolder}/configs/mocha.opts", "${file}"],
+      "args": ["--opts", "${workspaceFolder}/packages/client/mocha.opts", "${file}"],
       "console": "integratedTerminal",
-      "internalConsoleOptions": "neverOpen"
+      "internalConsoleOptions": "neverOpen",
+      "env": {
+        "TS_NODE_PROJECT": "${workspaceFolder}/packages/client/tsconfig.json"
+      }
     },
     {
       "type": "node",
@@ -17,9 +20,12 @@
       "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
       "args": [
         "--opts",
-        "${workspaceFolder}/configs/mocha.opts",
-        "${workspaceFolder}/src/**/*.spec.ts"
+        "${workspaceFolder}/packages/client/mocha.opts",
+        "${workspaceFolder}/packages/client/src/**/*.spec.ts"
       ],
+      "env": {
+        "TS_NODE_PROJECT": "${workspaceFolder}/packages/client/tsconfig.json"
+      },
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen"
     }
--- a/packages/client/src/features/tools/marquee-behavior.ts
+++ b/packages/client/src/features/tools/marquee-behavior.ts
@@ -15,7 +15,7 @@
  ********************************************************************************/
 
 import { Bounds, Point } from '@eclipse-glsp/protocol';
-import { injectable } from 'inversify';
+import { PointToPointLine } from 'sprotty';
 import { DrawMarqueeAction } from '../tool-feedback/marquee-tool-feedback';
 
 export interface IMarqueeBehavior {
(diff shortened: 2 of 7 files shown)


planger:
Thank you very much @ivy-lli! This looks great!
I'll ask @sgraband to review as he is most familiar with the original implementation.

.vscode/launch.json
```diff
@@ -6,9 +6,12 @@
...
       "name": "Run current test",
       "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
-      "args": ["--opts", "${workspaceFolder}/configs/mocha.opts", "${file}"],
+      "args": ["--opts", "${workspaceFolder}/packages/client/mocha.opts", "${file}"],
       "console": "integratedTerminal",
-      "internalConsoleOptions": "neverOpen"
+      "internalConsoleOptions": "neverOpen",
+      "env": {
+        "TS_NODE_PROJECT": "${workspaceFolder}/packages/client/tsconfig.json"
+      }
```
sgraband, line 14: Could you explain, what we need this for?

packages/client/src/features/tools/marquee-behavior.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    private linesIntersect(p1: Point, p2: Point, p3: Point, p4: Point): boolean {
+        const tCount = (p1.x - p3.x) * (p3.y - p4.y) - (p1.y - p3.y) * (p3.x - p4.x);
+        const tDenom = (p1.x - p2.x) * (p3.y - p4.y) - (p1.y - p2.y) * (p3.x - p4.x);
+        const t = tCount / tDenom;
+        const uCount = (p2.x - p1.x) * (p1.y - p3.y) - (p2.y - p1.y) * (p1.x - p3.x);
+        const uDenom = (p1.x - p2.x) * (p3.y - p4.y) - (p1.y - p2.y) * (p3.x - p4.x);
+        const u = uCount / uDenom;
+        if (t >= 0.0 && t <= 1.0 && u >= 0.0 && u <= 1.0) {
```
sgraband, line 116: We could reuse the `isBetween` method here right?

sgraband (changes requested):
Thank you @ivy-lli! The changes work great :slightly_smiling_face:. I just have some minor comments on the code.

packages/client/src/features/tools/marquee-behavior.ts
planger: I remember having written something similar in Sprotty directly:
https://github.com/eclipse/sprotty/blob/139fce21baab249b3ec9bc9e6d519558b876a8f9/packages/sprotty/src/utils/geometry.ts#L422

Maybe we can re-use parts of Sprotty's geometry utils.

packages/client/src/features/tools/marquee-behavior.ts
```diff
@@ -0,0 +1,165 @@
...
+export interface IMarqueeBehavior {
+    readonly entireElement: boolean;
+    readonly entireEdge: boolean;
+}
+
+@injectable()
+export class TouchMarqueeBehavior implements IMarqueeBehavior {
+    entireEdge = false;
+    entireElement = false;
+}
```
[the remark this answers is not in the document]
ivy-lli: Yes this is the default behavior. 
It's okay for me to adjust this as you suggest.

.vscode/launch.json
ivy-lli: I'm not really sure if this is really needed, I copied it from our own glsp integration, because I couldn't start the mocha tests over the launch config. But maybe the right path to the mocha config file is enough. I'll check that.

packages/client/src/features/tools/marquee-behavior.ts
ivy-lli: I'll check that 🙂

.vscode/launch.json
ivy-lli: If I try to run the test launch configs without this hint, I get a lot of errors in the log:
image

sgraband: I just now realized, that we are using mocha and not jest. Seems like setting the env variable is the way to go here. Please disregard my comment :slightly_smiling_face:

packages/client/src/features/tools/marquee-behavior.ts
ivy-lli: I could use the PointToPointLine#intersection method instead of our own linesIntersect method.

ivy-lli:
Hi @sgraband, @planger 
I made some adjustments as you suggested. Please check it again if you find a free minute.

sgraband (approved):
Thanks a lot @ivy-lli! Works great and LGTM :+1:

PR #157, opened 2021-12-08
470: Remove custom LayoutRegistry and reuse the one from Sprotty

CamilleLetavernier:
Note: there is one minor difference between the Sprotty registry and the one introduced in #470. When multiple layouts are bound to the same key, Sprotty logs a warning and rejects the new one; in the custom registry, since default layouts were not registered via dependency injection, we allowed overriding them (also after logging a warning).

We now have the expected Sprotty behavior: registering 2 different layouts on the same key is not allowed, and the first one wins. We must use rebind() instead.


tortmayr (approved):
Thanks. LGTM!

PR #158, opened 2021-12-10
224: FeedbackEdge uses wrong child when used inside container element

CamilleLetavernier:
- Fix anchor computation for nested elements

Unrelated: remove the debug-attribute on Category compartments (Blue border around the compartment)

fixes eclipse-glsp/glsp/issues/224

### DIFF
--- a/packages/client/src/features/tool-feedback/creation-tool-feedback.ts
+++ b/packages/client/src/features/tool-feedback/creation-tool-feedback.ts
@@ -38,7 +38,7 @@ import {
     SRoutableElement,
     TYPES
 } from 'sprotty';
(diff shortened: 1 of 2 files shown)


CamilleLetavernier:
Note: I introduced 2 new functions (absoluteToLocal/absoluteToParent). I suspect we'll need to have them in a more accessible place quite soon, as they might be useful for all operations that involve nested elements (Which should be... pretty much all graphical operations at some point :) ). But for now, they're only used for the Anchor computation, so I'll move them when necessary (I'm not sure where exactly, yet :) ).

tortmayr (approved):
LGTM. Thanks! :+1:

PR #160, opened 2022-01-07
Improve styling and simplfy workflow example

planger:
https://github.com/eclipse-glsp/glsp/issues/492

### DIFF
--- a/packages/client/css/glsp-sprotty.css
+++ b/packages/client/css/glsp-sprotty.css
@@ -52,17 +52,17 @@
 .sprotty-resize-handle {
     /* radius does not work in Firefox */
     r: 5px;
-    fill: #884;
+    fill: #1d80d1;
     stroke: none;
     z-index: 1000;
 }
 
 .sprotty-resize-handle.selected {
-    fill: rgb(41, 43, 46);
+    fill: #1d80d1;
 }
 
 .sprotty-resize-handle.active {
-    fill: rgb(41, 43, 46);
+    fill: #1d80d1;
 }
 
 .sprotty-edge {
@@ -89,7 +89,7 @@
 
 .sprotty-node.selected,
 .sprotty-edge.selected {
-    stroke: orange;
+    stroke: #1d80d1;
     stroke-width: 1.5px;
     stroke-dashoffset: 5;
     stroke-dasharray: 5, 5;
@@ -154,7 +154,7 @@
 
 .sprotty-edge > .sprotty-routing-handle {
     r: 5px;
-    fill: #884;
+    fill: #1d80d1;
     stroke: none;
     z-index: 1000;
 }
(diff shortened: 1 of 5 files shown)


packages/client/css/glsp-sprotty.css
```diff
@@ -50,23 +51,59 @@
...
+.sprotty-node:not(.selected) {
+    stroke-width: 0px;
+}
+
+.sprotty-node.mouseover:not(.selected),
+.sprotty-edge.mouseover:not(.selected) {
+    stroke: #1d1e1f;
+    stroke-width: 1.5px;
+    stroke-dashoffset: 5;
+    stroke-dasharray: 5, 5;
```
ndoschek, line 90: In general, I like the visual differentiation for selection/mouseover via dashed stroke. But due to the dashed line, edges might flicker on mouseover (see also screenshot). Maybe we could think of a solid line in a contrast color?

glsp-example-dashed-flickering

ndoschek (commented):
Thanks Philip! I like the improved styling of the workflow example! :tada: 
I spotted one potential issue with the dashed strokes, please see inline.

packages/client/css/glsp-sprotty.css
planger: Thanks Nina, you are right! We typically fix those issues by adding a transparent wider edge above the visible edge, which helps selection the edge too.
However, we also want to stay very generic, so probably dashes aren't a good choice for hover feedback.
I replaced it with opacity for now.
What do you think?
Also feel free to suggest different colors for the nodes if you don't like them.

ndoschek (changes requested):
Thanks Philip! The selection using opacity looks great!

I played around with the colors a bit, and using complementary colors for the nodes would look great to me, e.g. some darker orange (e.g. #db8651) and some teal (e.g. #5b9fa8).
This would also have the benefit that the blue selection color would give bit more contrast to both colors imo.
Also, I would suggest to use `var(--theia-focusBorder)` instead of `var(--theia-menu-selectionBackground)` as selection color, as it a bit brighter in dark mode as well.

Proposal in light theme:
image

Proposal in dark theme:
image

Also, I noticed two other issues:
- The weighted edges are not colored differently anymore
- The styling of the workflow standalone example: i.e. the resize handles differ from the selection color. Should we align that as well?

image

planger:
Thanks @ndoschek!
I used your suggested colors, added different color for weighted edges, tested for color blindness (see below), used theia color for selection and handles in standalone.

Red-Weak/Protanomaly:
image

I think it still is nicely distinguishable.

ndoschek (changes requested):
Thanks a lot Philip, looks great to me! :tada: 

Could you please update the file headers with the current year here as well, thanks!

planger:
Sure, added 4495c41. Thanks!

ndoschek (approved):

PR #167, opened 2022-03-07
#559 Update to Typescript 4.5.5

tortmayr:
- Consume new @eclipse-glsp/config version and update to typescript 4.5.5.
- Adapt code to conform to new "strict" mode and "noImplicitOverride"
- Cleanup dev dependencies and move common dependencies into package root

Also:
- Clean up and refactor `GLSPMouseTool`
- Clean up and document `smodel-util` and `array-util` utility  functions.

Part of eclipse-glsp/glsp/issues/559


tortmayr:
@martin-fleck-at As part of this change I also cleaned up the `RankingMouseTool`. Could you maybe have a special look at this?
I removed the overriding of the `decorate` method because it seemed to be unnecessary and as far as I can see everything works as expected. I know it was a long time ago but you initially implemented the `RankingMouseTool` (https://github.com/eclipsesource/graphical-lsp/blame/master/client/packages/sprotty-client/src/features/mouse-tool/mouse-tool.ts). Do you maybe recall why  overriding `decorate` was necessary in the first place?

martin-fleck-at:
> @martin-fleck-at As part of this change I also cleaned up the `RankingMouseTool`. Could you maybe have a special look at this? I removed the overriding of the `decorate` method because it seemed to be unnecessary and as far as I can see everything works as expected. I know it was a long time ago but you initially implemented the `RankingMouseTool` (https://github.com/eclipsesource/graphical-lsp/blame/master/client/packages/sprotty-client/src/features/mouse-tool/mouse-tool.ts). Do you maybe recall why overriding `decorate` was necessary in the first place?

@tortmayr I could not find any differences between the latest sprotty mouse tool and this one. So I think we can remove the override if everything still works as expected. Thank you for following up on this!

packages/client/src/features/mouse-tool/mouse-tool.ts
```diff
@@ -24,48 +23,35 @@ export interface IMouseTool {
     deregister(mouseListener: MouseListener): void;
 }
 
+/**  Custom helper type to declare the explicit mouse listener methods
```
martin-fleck-at, line 26: Odd formatting.

packages/client/src/features/mouse-tool/mouse-tool.ts
```diff
@@ -85,15 +71,13 @@ export class RankingMouseTool extends MouseTool implements IMouseTool {
...
+    async dispatchActions<K extends MouseListenerMethods>(
         mouseListeners: MouseListener[],
         methodName: K,
         element: SModelElement,
         event: MouseEvent
     ): Promise<void> {
-        const actions = mouseListeners
-            .map(listener => listener[methodName].apply(listener, [element, event]))
-            .reduce((a, b) => a.concat(b));
+        const actions = mouseListeners.map(listener => listener[methodName](element, event as WheelEvent)).reduce((a, b) => a.concat(b));
```
martin-fleck-at, line 80: Could you explain why the event is cast to `WheelEvent`? I believe this method is generic for any MouseEvent but the cast also does not seem to break anything. In any case, if you want to leave it, please explain why this is happening.

packages/client/mocha.opts
```diff
@@ -1,7 +0,0 @@
---require ts-node/register
```
martin-fleck-at, line 1: With or even without this file, I cannot run the tests from the launch.json `Run current test` or `Run all tests`. Even before any execution I get:
```
✖ ERROR: --opts: configuring Mocha via 'mocha.opts' is DEPRECATED and no longer supported.
          Please use a configuration file instead.         
```

So it would be great if you could either remove the two entries from launch.json since the file is no longer there anyway or update the calls accordingly.

martin-fleck-at (changes requested):
Hello @tortmayr! Thank you very much for your change. I believe everything still works as expected, I just have a few comments which may need to be addressed. Other than that, great work and finally we get to use `override` :partying_face:

packages/client/src/features/mouse-tool/mouse-tool.ts
tortmayr: Sure: the sprotty `MouseListener` as a `wheel` method which expects a `WheelEvent` as parameter. 
So due to new strict typescript configuration the compiler would complain if  we wouldn't cast the event. 
However casting is totally safe in this case because `WheelEvent` extends `MouseEvent` and can therefore also be safely used in all other listener methods that just expect a `MouseEvent`.
The original sprotty implementation handles this in the same fashion: 
https://github.com/eclipse/sprotty/blob/4c111d2a96b8e2bfef6acb290ec0d185ddb752c7/packages/sprotty/src/base/views/mouse-tool.ts#L66

packages/client/mocha.opts
tortmayr: Fixed

packages/client/src/features/mouse-tool/mouse-tool.ts
tortmayr: Fixed

packages/client/src/features/mouse-tool/mouse-tool.ts
martin-fleck-at: I see, thank you for the explanation. I saw that it was not causing problems but still it looked odd. If sprotty does it in the same way we can definitely do the same.

martin-fleck-at (approved):
Thank you @tortmayr for the quick updates and explanation. Everything looks great now!

PR #168, opened 2022-03-16
Fix DefaultModelInitializationConstraint

ivy-lli:
Wait for SetModelAction instead of UpdateModelAction to mark model as initialized.


tortmayr (commented):
In general the changes look good to me. However, to be more resilient against future errors I would recommend to resolve the promise on both (non-empty) `SetModelAction`s and `UpdateModelAction`s. Sprotty has a special handling for `UpdateModelAction`s and they are treated like a `SetModelAction` if the current model root is empty. So both action types should resolve the initialization constraint.

tortmayr (approved):
LGTM :+1:  Thanks for your contribution!

PR #169, opened 2022-03-23
#602 Fix SelectionService.getSelectedElements() method

tortmayr:
With eclipse-glsp/glsp-client/pull/167 a bug in the selection service was introduced. The `getSelectedElements()`  query method  returns
all `selectable` elements instead of the elements that are actually selected.

- Fix `SelectionService.getSelectedElements()` behavior. This also includes a performance improvement. We now only map the currently selected element ids to their SModelElement Instead of iterating over all elements of the index to receive the selected elements
- Introduce  a new  smodel util function that allows retrieving all model elements that correspond to a given set of element ids from the index.
- Introduce a `TypeGuard` convenience type to `typeguard-util`

Fixes eclipse-glsp/glsp/issues/602

### DIFF
--- a/packages/client/src/features/select/selection-service.ts
+++ b/packages/client/src/features/select/selection-service.ts
@@ -31,7 +31,7 @@ import {
 } from 'sprotty';
 import { SModelRootListener } from '../../base/model/update-model-command';
 import { GLSP_TYPES } from '../../base/types';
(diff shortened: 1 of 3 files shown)


martin-fleck-at (approved):
Looks good to me, thank you Tobi!

PR #175, opened 2022-03-31
Enable default tools on focus loss

planger:
Fixes https://github.com/eclipse-glsp/glsp/issues/547

### DIFF
--- a/packages/client/src/base/container-modules.ts
+++ b/packages/client/src/base/container-modules.ts
@@ -44,7 +44,7 @@ import { navigationModule } from '../features/navigation/di.config';
 import glspSelectModule from '../features/select/di.config';
 import toolFeedbackModule from '../features/tool-feedback/di.config';
 import toolPaletteModule from '../features/tool-palette/di.config';
-import toolsModule from '../features/tools/di.config';
+import { enableDefaultToolsOnFocusLossModule, toolsModule } from '../features/tools/di.config';
 import { markerNavigatorModule, validationModule } from '../features/validation/di.config';
 import glspViewportModule from '../features/viewport/di.config';
 import defaultGLSPModule from './di.config';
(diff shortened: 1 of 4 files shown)


tortmayr (approved):
Seems to work great and code looks good to me.
However, I think we need to reconsider our customization strategy for DI modules.
I'm not the biggest fan of creating a lof of additional DI modules that contain only one binding just for the purpose of making it configurable.
For now its fine, but in the future I think we need a more refined approach.

planger:
Thanks for the feedback! I appended `Handler` to the class name.

> However, I think we need to reconsider our customization strategy for DI modules.
> I'm not the biggest fan of creating a lof of additional DI modules that contain only one binding just for the purpose of making it configurable.
> For now its fine, but in the future I think we need a more refined approach.

Yeah, I understand. I think it depends a bit whether we consider this to be an optional feature. If yes it should probably be in a separate module, because this makes it very easy to exclude it. Otherwise adopters would have to manually unbind/rebind all or specific bindings that concern the respective feature from the bindings of the module this feature resides in. This is very cumbersome and brittle (if something changes).

In this particular case I agree it is borderline. I'd consider it rather a default thing, but you made a good point that it [should be optional](https://github.com/eclipse-glsp/glsp/issues/547#issuecomment-1044350445). Therefore, I put it into a separate module.

Another option is to always bind it and keep it in a more general module, but introduce switches or configs that are bound by adopters into the di module, which activate or deactivate a certain feature. Could be an alternative in the future for things we typically see as default but want to make easy to deactivate. But it increases the burden of testing.

tortmayr:
> Another option is to always bind it and keep it in a more general module, but introduce switches or configs that are bound by adopters into the di module, which activate or deactivate a certain feature. Could be an alternative in the future for things we typically see as default but want to make easy to deactivate. But it increases the burden of testing.

Yeah I think the switches/configs routes is definitely worth investigating. This would also make sense for this feature. I would see it as the responsibility of the `ToolManager` to decide whether the default tools should be restored on focus loss. With switches/configs we could do that. Maybe a pattern similar to what sprotty is using for its `ViewerOptions`. (i.e. the `overwriteViewerOptions` function. But that's something we should discuss in the next dev meeting.

PR #177, opened 2022-04-08
Issue-613 Fix snap routing handles

ivy-lli:
Fix for: https://github.com/eclipse-glsp/glsp/issues/613

Also snap the routing handle before send it to the server.
With this change it will be possible to draw a straight line out of a polyline edge.


ivy-lli:
With fix:
glsp-waypoint-fix

tortmayr (approved):
Thanks for the fix! Works like a charm. And nice that you also added the possibility to use `unsnapped` routing with the alt key. :+1:

PR #178, opened 2022-04-26
Improve default configuration for foreign objects

tortmayr:
Disable `selectFeature` and `moveFeature` for the default foreign object configuration.
The select feature doesn't make sense for `ForeignObjectElements` as each DOMElement rendered from the objects `code` property
might provide it's own selection behavior. In addition this ensure that the resize feature also works for nodes that contain `ForeignObjectElements`.

Typically `ForeignObjectElement`s are children of a GNode so we also disable the move feature as nested child elements should not be movable.

Also: Add missing `Button` definition to default types and clean up imports of  `model.ts`


planger (approved):

PR #181, opened 2022-05-19
GH-652 Configurable gridsnapper

tortmayr:
Make grid size of the `GridSnapper` configurable via constructor and setter.
Fixes https://github.com/eclipse-glsp/glsp/issues/652


planger (approved):

PR #183, opened 2022-05-28
GH-628 Use generic context menu mouse event instead of right click event

tortmayr:
- Update to latest sprotty version
- Update `SelectionServiceAwareContextMenuMouseListener` to use the `contextmenu` mouse event instead of the `right-click` event.

Fixes https://github.com/eclipse-glsp/glsp/issues/628

### DIFF
--- a/packages/client/src/features/context-menu/selection-service-aware-context-menu-mouse-listener.ts
+++ b/packages/client/src/features/context-menu/selection-service-aware-context-menu-mouse-listener.ts
@@ -15,14 +15,7 @@
  ********************************************************************************/
 import { Action } from '@eclipse-glsp/protocol';
 import { inject, injectable, optional } from 'inversify';
-import {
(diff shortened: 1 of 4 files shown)


planger (approved):
Looks great, thanks! Now also the context-menu button on the keyboard works! :+1: 

However, I found an issue:
If you select an element first (left-click) and then right-click on another element, the previous element is not unselected. So if you now select "Delete" from the context menu, the previously selected element is also deleted.

Imho this is a bit unexpected.

I suggest that we now modify the select mouse tool to also update the selection on any other mouse click (not just mouse button 0 === left-click). This is actually more consistent with other software, if you e.g. try a right-click or even wheel click in VS Code you'll see that it updates the selection too). This makes also the special handling regarding the selection in the context menu listener obsolete.

I'll append a commit for you to review.

tortmayr:
Thanks. Looks good to me!

PR #186, opened 2022-06-29
Update workflow example to not use `GLSPProjectionView`

tortmayr:
Temporary remove usage of `GLSPProjectionView` in the workflow example. 
Can be readded once https://github.com/eclipse-glsp/glsp/issues/689 is resolved.


planger (approved):

PR #187, opened 2022-07-06
694: Improve Layouter to support more dynamic layouts and complex parent/children node structures

CamilleLetavernier:
This PR depends on https://github.com/eclipse-glsp/glsp-server/pull/175

- Introduce a new StatefulLayouter, which applies layout in 2 passes: children -> parent for the "minimal size" (resizeContainer), then parent -> children to take the correct parent size into account for alignment (alignment, vGrab/hGrab)
- Update HBox and VBox layouts to take the new layouter into account
- Do not inherit child-specific layout properties (PrefSize, Grab)
- Properly take padding into account
- Update the Style for nested categories, to make them more readable (Add a border, use a different color for nested categories)

fixes https://github.com/eclipse-glsp/glsp/issues/694
fixes https://github.com/eclipse-glsp/glsp/issues/610
fixes https://github.com/eclipse-glsp/glsp/issues/471

Contributed on behalf of STMicroelectronics.

### DIFF
--- a/examples/workflow-glsp/css/diagram.css
+++ b/examples/workflow-glsp/css/diagram.css
@@ -65,6 +65,13 @@
     fill: #5c87bd;
 }
 
+.category .category > .sprotty-node {
+    /** Add a border to nested categories, so we can distinguish them */
+    stroke: #38679a;
+    fill: #6f9ad0;
+    stroke-width: 1;
+}
+
 .sprotty-edge.weighted.low:not(.selected),
 .sprotty-edge.weighted.low:not(.selected) .arrow {
     stroke: rgb(128, 90, 233);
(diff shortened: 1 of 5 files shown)


CamilleLetavernier:
Dynamic layouts examples:

1. Structure compartment (hGrab=true, vGrab=false), Left-aligned label (hGrab=true), right-aligned label (hGrab=true). This is the main use case we wanted to support



2. Nested categories (new appearance to make them more readable. Before, they use the same color and didn't have a border, so overlapping categories were "invisible"). Other than the appearance, the behavior should be unchanged.



3. Other cases (simple tasks, either top level, or nested in categories). Behavior is mostly unchanged, although https://github.com/eclipse-glsp/glsp/issues/471 is now fixed.

CamilleLetavernier:
Example model (The top-left category is hand-crafted by editing the Json; you can't create it via normal GLSP edition):

[issue_694.zip](https://github.com/eclipse-glsp/glsp-client/files/9055562/issue_694.zip)

ndoschek (approved):
Thanks Camille, changes look great to me, different scenarios regarding the improved layout work as expected and described in the linked issues! Thanks again! :+1:

PR #189, opened 2022-07-14
Update download server script

tortmayr:
Ensure that the snapshot version of the example server is downloaded


ndoschek (approved):
Thanks Tobias, LGTM and works as expected :+1:

PR #191, opened 2022-07-20
Fix keylisteners when using 'GLSPProjectionView'

gabrieldulca:
Fixes the [Broken Keylisteners](https://github.com/eclipse-glsp/glsp/issues/689) when using the GLSPProjection view by making the container `div` focusable.


tortmayr (changes requested):
Looks good to me and seems to work well for the Theia integration.
However, the styling of standalone example is broken:
Screenshot from 2022-07-20 17-19-18

Could you please have a look at that? Thanks!

gabrieldulca:
Thanks for the feedback! The styling for the standalone example is now fixed.

tortmayr (approved):

PR #194, opened 2022-08-02
Refactor download-server script

tortmayr:
Refactor configuration for the download server-script into separate JSON resource. This makes it easier to maintain and offers a simple approach to update the configuration in automated (CI) jobs


planger (approved):

PR #196, opened 2022-08-14
#642: Fix export SVG handling

martin-fleck-at:
- SVG element is not properly detected as it is nested
- SVG element is not properly copied when it does not have an ID

https://github.com/eclipse-glsp/glsp/issues/642

### DIFF
--- a/packages/client/src/features/export/glsp-svg-exporter.ts
+++ b/packages/client/src/features/export/glsp-svg-exporter.ts
@@ -1,0 +1,46 @@
+/********************************************************************************
+ * Copyright (c) 2022 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


planger (commented):
Thank you very much, @martin-fleck-at! Works great! Except when I open the exported SVG in a browser. It then seems like the SVG has a wrong size (super tiny) and thus isn't visible. Do you have the same issue? Is there an easy fix for that?

The main reason I'm asking is because some of the GLSP adopters use HTML foreign elements, which only work in browsers. So if they want to show their export, they'd have to use the browser, which doesn't seem to work though.

planger (approved):
Excellent, thanks!

> Please note, however, that we do not apply any scaling so the image will always be the same size independent from the current scaling in the editor.

Yeah, that's clear. I would not have expected that since it is rather an editor state and the export should be stateless.

Thanks!

PR #198, opened 2022-09-05
#612: Recalculate all routing points on routing point update

martin-fleck-at:
Fixes https://github.com/eclipse-glsp/glsp/issues/612

### DIFF
--- a/packages/client/src/features/tools/change-bounds-tool.ts
+++ b/packages/client/src/features/tools/change-bounds-tool.ts
@@ -16,9 +16,7 @@
 import {
     Action,
     Bounds,
-    ChangeBoundsOperation,
-    ChangeRoutingPointsOperation,
-    CompoundOperation,
+    ChangeBoundsOperation, CompoundOperation,
     Dimension,
     ElementAndBounds,
     ElementAndRoutingPoints,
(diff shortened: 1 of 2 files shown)


packages/client/src/features/tools/change-bounds-tool.ts
```diff
@@ -246,9 +244,6 @@ export class ChangeBoundsListener extends DragAwareMouseListener implements Sele
                 element.outgoingEdges.map(toElementAndRoutingPoints).forEach(ear => newRoutingPoints.push(ear));
             }
         });
-        if (newRoutingPoints.length > 0) {
-            result.push(ChangeRoutingPointsOperation.create(newRoutingPoints));
```
tortmayr, line 250: Removing this introduces a regression bug.
If an node is moved and the connected edges a rerouted in the process the changed routing points are no longer sent to  the server. As a consequence an new route based on the old routingpoints is computed after the server update.
(This bug is not visible with polyline routing, you would have to use an manhattan edge to test this)

However, we should probably reuse the update method that you introduced with this PR.

tortmayr (changes requested):
Thanks Martin. In general, the change works well, however, I noticed one regression.

packages/client/src/features/tools/change-bounds-tool.ts
martin-fleck-at: Good catch! I extracted the calculation of the routing points and now share the functionality between both tools. Thank you!

tortmayr (approved):
Thanks! LGTM 👍🏼

PR #209, opened 2022-12-16
fix(diagram): scope css to avoid conflict on existing layout

jmfrancois:
Using glsp client on a custom layout is broken because of the global css applied to `svg`. all my application icons get a border.

The result I get from the standalone example
image

As this is my first contribution to the project please tell me if you have any CI for the changelog or if I need to update it by hand. (I am used to changeset but it is used here)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changes
 
 -   [layout] Improve Layouter to support more dynamic layouts and complex parent/children node structures [#187](https://github.com/eclipse-glsp/glsp-client/pull/187) - Contributed on behalf of STMicroelectronics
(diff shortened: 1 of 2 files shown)


planger:
@jmfrancois Thank you very much for your contribution! Overall this change looks fine, we just need to validate whether follow-up changes are needed in the platform integrations (Theia, VS Code) due to some CSS specificity changes caused by this change.

I'll test that next week and get back to you.

jmfrancois:
Yes for sure it may brake some specific integration due to this change. 
Thank you for your feedback !

planger (approved):
I've tested this change with the Theia integration and didn't see any issues caused by scoping the generic Sprotty rules. So I'm fine merging this PR.

Thanks again for your contribution @jmfrancois!

PR #210, opened 2022-12-16
Update to latest vscode-json-rpc version

martin-fleck-at:


tortmayr (approved):

tortmayr:
Had to be reverted because of a node version comp-ability (vscode-ws-jsonrpc requires node 16, our current minimal verison is node 14). We have to discuss how to continue with this issue

PR #212, opened 2022-12-17
Fix copy&paste support for GLSPProjectionView

tortmayr:
The `isDiagramActive` method of the copy-paste handler did check whether the active element is an instance of svg. 
In the case of the `GLSPProjectionView` this check failed because the SVG is contained by a wrapper container.
We can omit this part of the check entirely because we already know that the diagram is active if the parent of the active element is the `baseDiv`

Fixes https://github.com/eclipse-glsp/glsp/issues/827

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -43,7 +43,7 @@
   },
   "scripts": {
     "build": "tsc -b",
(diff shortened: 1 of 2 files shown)


planger (approved):

PR #214, opened 2022-12-19
Integrate coverage reports

tortmayr:
- Update config dev packages and add build scripts for coverage reports
- Fix launch configurations for testing
- Fix error in workflow standalone. (vscode-ws-jsonrpc is now a pure es module and has to be imported dynamically)
- Integrate coverage reporting  into Jenkinsfile Part of https://github.com/eclipse-glsp/glsp/issues/820


martin-fleck-at (approved):
LGTM!

PR #215, opened 2022-12-21
Provide own websocket vsode-jsonrpc connection

tortmayr:
Replace vscode-ws-jsonrpc with own implementation

Part of https://github.com/eclipse-glsp/glsp/issues/944


martin-fleck-at:
@tortmayr I think this is a good idea! I was also surprised about all the necessary changes to use vscode-ws-jsonrpc. But since they dropped the ESM support, I think it makes sense to maintain the code ourselves to not pollute our code with dynamic imports. The layer is thin enough and will probably not change too often anyway. Thank you Tobias!

martin-fleck-at (approved):
LGTM!

PR #217, opened 2022-12-30
Ensure that startup UI extensions are rendered correctly

tortmayr:
UI extensions that should be active i.e. visible on startup might be activated to early (before the initial model loading is completed) which can cause rendering issues in some cases. This can be avoided by enabling these UI extensions after the model has been initialized. (see also https://github.com/eclipse-glsp/glsp/issues/585)

For the ToolPalette in particular this issue is not noticeable in the workflow diagram because the extension is activated after it has retrieved all tool palette items from the server. This typically takes longer than the initial model loading. Nevertheless we should simply establish the best practice of always enabling UI extensions after the model initialization.


planger (approved):

PR #218, opened 2023-01-02
Cleanup diagram modules

tortmayr:
- Remove interdependency between exportSvg module and toolsModule by creating a `glspExportSvgModule`
- Use `glspExportSvgModule` as drop-in replacement for sprotty's exportSvgModule (DEFAULT_MODULES)
- Make `DEFAULT_MODULES` a readonly tuple instead of an modifiable array (as const)
- Introduce new `createDiagramContainer` utility function and deprecate `createClientContainer` function We typically refer to the sprotty DI container as `diagram container`, so the utility function should also reflect this naming convention. In addition, exclude descriptions can now be passed as optional arguments. This allows to dynamically exclude a default module before loading it in the container. This avoids the overhead of (unnecessary loading and )manually unloading a default module after the container creation.

### DIFF
--- a/packages/client/src/base/container-modules.ts
+++ b/packages/client/src/base/container-modules.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2021 EclipseSource and others.
+ * Copyright (c) 2021-2023 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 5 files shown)


packages/client/src/base/container-modules.ts
```diff
@@ -83,7 +84,52 @@ export const DEFAULT_MODULES = [
...
+ *
+ * ```typescript
+ * rebind(NavigationTargetResolver).to(MyNavigationTargetResolver);
+ * ```
+ *
+ * @param customModules Custom modules to be loaded in addition to the default modules and/or default modules that should be excluded.
+ * @param options Options to customize the module loading
+ * @returns The created container.
+ */
+
```
martin-fleck-at, line 111:
```suggestion
```

Remove unnecessary empty line.

martin-fleck-at (approved):
Everything works and the custom svg export handling is still active, thank you very much Tobias! I just found one very minor, unnecessary empty line that could be removed.

PR #221, opened 2023-01-12
Add custom 'GEdge' that provides a 'localToParent' calculation

martin-fleck-at:
Fixes https://github.com/eclipse-glsp/glsp/issues/855

fix-hover

So far I did not detect any side effects. If that keeps being true, we should contribute this back to Sprotty as well.


tortmayr (approved):
LGTM!

PR #222, opened 2023-01-18
Add common Disposable API

tortmayr:
Provide a reusable Disposable API similar to what Theia/VS Code uses
Fixes https://github.com/eclipse-glsp/glsp/issues/867

Also: Update dev packages and autoformat package.json files


packages/protocol/src/utils/disposable.ts
```diff
@@ -0,0 +1,95 @@
...
+export interface Disposable {
+    /**
+     * Dispose this object.
+     */
+    dispose(): void;
+}
+
+export namespace Disposable {
+    export function is(value: unknown): value is Disposable {
+        return AnyObject.is(value) && hasFunctionProp(value, '');
```
martin-fleck-at, line 31:
```suggestion
        return AnyObject.is(value) && hasFunctionProp(value, 'dispose');
```

Some test cases:
```
    describe('is', () => {
        it('primitive', () => {
            expect(Disposable.is('A')).to.be.false;
        });
        it('any function', () => {
            expect(Disposable.is('A'.toString)).to.be.false;
        });
        it('create', () => {
            expect(Disposable.is(Disposable.create(() => 'A'.toString()))).to.be.true;
        });
        it('empty', () => {
            expect(Disposable.is(Disposable.empty())).to.be.true;
        });
        it('any object', () => {
            const obj = { a: 1, b: 2 };
            expect(Disposable.is(obj)).to.be.false;
        });
        it('disposable object', () => {
            const obj: Disposable = { dispose: () => 'ok' };
            expect(Disposable.is(obj)).to.be.true;
        });
    });
```

martin-fleck-at (changes requested):
API looks good to me in general, thank you very much Tobias! If we ever need more functionality (e.g., listeners) we can always add them later. I was only wondering on whether we should already use the API in this package, e.g., there are a few places where we have `register`/`deregister` methods (e.g., tools, listeners, etc.) where the `register` call could return the corresponding deregister instead of void. Or will there be a separate task/PR for that?

tortmayr:
> If we ever need more functionality (e.g., listeners) we can always add them later. I was only wondering on whether we should already use the API in this package, e.g., there are a few places where we have `register`/`deregister` methods (e.g., tools, listeners, etc.) where the `register` call could return the corresponding deregister instead of void. Or will there be a separate task/PR for that?

I think we should do that in a separate PR.

martin-fleck-at (approved):
LGTM!

PR #224, opened 2023-01-24
Ensure edge routing point can be grabbed in Firefox Browser

martin-fleck-at:
It seems that in Firefox the 'mouseEnter' event is fired directly after the 'mouseDown' event which caused us to reset the position. However, we should only reset the position if no moving is in progress (i.e., there is no current position tracking/dragging going on). But in that case there is no need to reset the position at all, therefore we can safely remove the 'mouseEnter' event handling for edge routing editing.

Fixes https://github.com/eclipse-glsp/glsp/issues/877


tortmayr (approved):
LGTM!

PR #228, opened 2023-02-08
Extend type-util

tortmayr:
Extends property check utility functions to also support checking of optional properties.


martin-fleck-at (approved):
LGTM!

PR #230, opened 2023-02-20
Ensure that drawFeedbackAction uses correct schema type

tortmayr:
and allow partial schemas
Fixes https://github.com/eclipse-glsp/glsp/issues/926


martin-fleck-at (approved):
LGTM!

PR #232, opened 2023-03-08
Fix definition  of `LayoutOperation`

tortmayr:
Align definition with sprotty's definition which also allows optional elementIds.

@xai Please also adpat in https://github.com/eclipse-glsp/glsp-website-source/pull/81


xai (approved):
LGTM :+1:

PR #234, opened 2023-03-12
Reformat code & overall cleanup

tortmayr:
- Reformat code with prettier
- Fix typos
- Ensure that everything is exported via root index
- Rework `GLSPCommandStack`. The undo/redo methods of the command stack should never be called as the server is responsible for handling those requests. In any case we now ensure that a proper warning is logged and the model remains unchanged if the methods are called on accident
- Remove unused `IGLSPToolManager` interface
- Move glsp layouter (vbox,hbxo,freeform) into proper module to avoid unnecessary rebinding
- Drop support for stack layout ( Fixes https://github.com/eclipse-glsp/glsp/issues/444)


examples/workflow-standalone/app/diagram.html
```diff
@@ -1,38 +1,34 @@
...
+            height: 100%;
+            overflow: hidden;
+        }
 
-            svg {
-                width: 100%;
-                height: 100%;
-            }
-        </style>
-        <!-- support Microsoft browsers -->
```
martin-fleck-at, line 28: Was this removed on purpose or formatted away automatically? I wasn't able to test this on Ubuntu but I just wanted to make sure this is what we ant to do.

packages/client/src/features/bounds/di.config.ts
```diff
@@ -15,12 +15,21 @@
...
+    configureLayout,
+    HBoxLayouter,
+    HiddenBoundsUpdater,
+    Layouter,
+    LayoutRegistry,
+    RequestBoundsCommand,
+    SetBoundsCommand,
+    VBoxLayouter
 } from 'sprotty';
-import '../../../css/command-palette.css';
```
martin-fleck-at, line 21: Good catch!

martin-fleck-at (approved):
Looks good to me, I just have one minor question regarding the removal of the "Microsoft browser support". Other than that, I'm very happy with the change and surprised we had that many undiscovered typos so thank you Tobi!

examples/workflow-standalone/app/diagram.html
tortmayr: Yes this was removed on purpose. It imported a non existing script ("lib/dom.js") which was needed previously to support Internet Explorer and earlier Versions of Edge but we removed the script a while back when IE reached EOL.

PR #236, opened 2023-03-16
GLSP-954 Add DI utility to protocol package

tortmayr:
- Create a `di-util` module that provides
  - A configurable function for initializing new DI containers with a specific module configuration (`initializeDiContainer`)
  - The `bindAsService` helper function which can be used as shorthand for the common pattern of binding classes to self before binding them to another service
  - The `bindOrRebind` helper function
- Double check all multiinjected service bindings and ensure that they can be easily rebound if necessary. This is achieved with the new `bindAsService` helper function

Part of https://github.com/eclipse-glsp/glsp/issues/954


packages/protocol/src/utils/di-util.ts
```diff
@@ -0,0 +1,166 @@
...
+ * Checks wether the given service identifier is already bound in the given context
+ * then either calls  the `bind` or `rebind` function respectively.
+ *
+ * As this is just a convenience function
+ * ```ts
+ * bindOrRebind({bind,isBound,rebind}, MyService).to(SomeOtherService);
+ * ```
+ * is equivalent to:
+ * ```
+ * if (isBound(MyService)){
```
martin-fleck-at, line 94:
```suggestion
 * if (isBound(MyService)) {
```

packages/protocol/src/utils/di-util.ts
```diff
@@ -0,0 +1,166 @@
...
+ *    rebind(MyService).to(SomeOtherService);
+ * }else {
```
martin-fleck-at, line 96:
```suggestion
 * } else {
```

martin-fleck-at (approved):
Change looks good to me and works well, very cool utility functions that I'm already happy to apply in other projects! I have a few comments but they should not stop this PR from being merged. Thank you, Tobias!

PR #237, opened 2023-03-17
GLSP-954 Ensure that bindAsAService also work with the DI container as context

tortmayr:
The `bindAsService` method failed when the DI container was used as context. To avoid this we have to make sure that the bind function receives the correct `this` arg.

Use utility functions in standalone di config


tortmayr:
For reference: adding this binding
```ts
bindAsService(container, TYPES.ModelSource, GLSPDiagramServer);
```
in master failes, but with this PR it works as expected

martin-fleck-at (approved):
LGTM!

PR #240, opened 2023-04-19
Add keyboard support for resizing elements

aylin-sarioglu:
This PR is part of my diploma research conducted at the [Model Engineering lab](https://me.big.tuwien.ac.at/) of TU Wien in the [Business Informatics Group](https://www.big.tuwien.ac.at/). It introduces the **resize** functionality.

## Example

1. Select one or multiple elements
2. Use the shortcut  `ALT + A` to activate the resize mode
3. Use the keys  `+` or  `-` to resize the element accordingly
4. Use `CTRL + 0` to reset to default size 

## Hints

1. The shortcut `ALT + A` is necessary, as later, `+` or  `-` , will also be used for other functionalities (e.g. zoom).
2. In a different PR, a user notification will be displayed if the mode is activated.

@martin-fleck-at, @ndoschek

Part of https://github.com/eclipse-glsp/glsp/issues/1029

### DIFF
--- a/packages/client/src/features/accessibility/resize-key-tool/di.config.ts
+++ b/packages/client/src/features/accessibility/resize-key-tool/di.config.ts
@@ -14,19 +14,24 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
+import { bindAsService, BindingContext } from '@eclipse-glsp/protocol';
 import { ContainerModule } from 'inversify';
-import { configureMoveZoom } from './move-zoom/di.config';
-import { configureResizeTools } from './resize-key-tool/di.config';
-import { configureSearchPaletteModule } from './search/di.config';
-import { configureViewKeyTools } from './view-key-tools/di.config';
+import { configureActionHandler } from 'sprotty';
+import { TYPES } from '../../../base/types';
+import { ResizeElementAction, ResizeElementHandler } from './resize-key-handler';
+import { ResizeKeyTool } from './resize-key-tool';
 
 /**
- * Enables the accessibility tools for a keyboard-only-usage
+ * Handles resize actions.
  */
-export const glspAccessibilityModule = new ContainerModule((bind, unbind, isBound, rebind) => {
+export const glspResizeKeyModule = new ContainerModule((bind, unbind, isBound, rebind) => {
     const context = { bind, unbind, isBound, rebind };
     configureResizeTools(context);
-    configureViewKeyTools(context);
-    configureMoveZoom(context);
-    configureSearchPaletteModule(context);
 });
+
+export function configureResizeTools(context: BindingContext): void {
+    context.bind(ResizeElementHandler).toSelf().inSingletonScope();
+
+    configureActionHandler(context, ResizeElementAction.KIND, ResizeElementHandler);
+    bindAsService(context, TYPES.IDefaultTool, ResizeKeyTool);
+}
--- a/packages/client/src/features/accessibility/resize-key-tool/resize-key-handler.ts
+++ b/packages/client/src/features/accessibility/resize-key-tool/resize-key-handler.ts
@@ -1,0 +1,120 @@
+/********************************************************************************
+ * Copyright (c) 2023 Business Informatics Group (TU Wien) and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { Action, ChangeBoundsOperation, Dimension, Point } from '@eclipse-glsp/protocol';
+import { inject, injectable, optional } from 'inversify';
+import { IActionDispatcher, IActionHandler, ICommand, ISnapper, SModelElement, SParentElement } from 'sprotty';
+import { EditorContextService } from '../../../base/editor-context-service';
+import { TYPES } from '../../../base/types';
+import { isValidMove, isValidSize, minHeight, minWidth } from '../../../utils/layout-utils';
+import { getElements, isSelectableAndBoundsAware, SelectableBoundsAware, toElementAndBounds } from '../../../utils/smodel-util';
+import { Resizable } from '../../change-bounds/model';
+import { GridSnapper } from '../../change-bounds/snap';
+
+// eslint-disable-next-line no-shadow
+export enum ResizeType {
+    Increase,
+    Decrease,
+    MinSize
+}
+
+/**
+ * Action for triggering resizing of elements.
+ */
+export interface ResizeElementAction extends Action {
+    kind: typeof ResizeElementAction.KIND;
+    elementIds: string[];
+    resizeType: ResizeType;
+}
+export namespace ResizeElementAction {
+    export const KIND = 'resizeElementAction';
+
+    export function is(object: any): object is ResizeElementAction {
+        return Action.hasKind(object, KIND);
+    }
+
+    export function create(elementIds: string[], resizeType: ResizeType): ResizeElementAction {
+        return { kind: KIND, elementIds, resizeType };
+    }
+}
+
+/* The ResizeElementHandler class is an implementation of the IActionHandler interface that handles
+resizing of elements. */
+@injectable()
+export class ResizeElementHandler implements IActionHandler {
+    @inject(EditorContextService)
+    protected editorContextService: EditorContextService;
+
+    @inject(TYPES.IActionDispatcher) protected dispatcher: IActionDispatcher;
+
+    // Default x resize used if GridSnapper is not provided
+    static readonly defaultResizeX = 20;
+
+    // Default y resize used if GridSnapper is not provided
+    static readonly defaultResizeY = 20;
+    protected grid = { x: ResizeElementHandler.defaultResizeX, y: ResizeElementHandler.defaultResizeY };
+
+    protected isEditMode = false;
+
+    constructor(@inject(TYPES.ISnapper) @optional() protected readonly snapper?: ISnapper) {
+        if (snapper instanceof GridSnapper) {
+            this.grid = snapper.grid;
+        }
+    }
+
+    handle(action: Action): void | Action | ICommand {
+        if (ResizeElementAction.is(action)) {
+            this.handleResizeElement(action);
+        }
+    }
+
+    handleResizeElement(action: ResizeElementAction): void {
+        const elements = getElements(this.editorContextService.modelRoot.index, action.elementIds, isSelectableAndBoundsAware);
+        this.dispatcher.dispatchAll(this.resize(elements, action));
+    }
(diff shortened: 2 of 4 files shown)


packages/client/src/features/change-bounds/di.config.ts
```diff
@@ -0,0 +1,25 @@
+/********************************************************************************
+ * Copyright (c) 2019-2023 EclipseSource and others.
```
martin-fleck-at, line 2:
```suggestion
 * Copyright (c) 2023 EclipseSource and others.
```

Please have a look at all the copyright headers and coordinate with Haydar on the correct copyright holder that should be applied.

packages/client/src/features/change-bounds/resize-tool.ts
```diff
@@ -0,0 +1,157 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        return actions;
+    }
+
+    protected handleResizeElement(element: SModelElement & BoundsAware, deltaWidth: number, deltaHeight: number): Action | undefined {
```
martin-fleck-at, line 120:
```suggestion
    protected handleResizeElement(element: SelectableBoundsAware, deltaWidth: number, deltaHeight: number): Action | undefined {
```

I think it would be nice to use the `SelectableBoundsAware` everywhere instead of the combined type.

packages/client/src/features/change-bounds/resize-tool.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    override keyDown(element: SModelElement, event: KeyboardEvent): Action[] {
+        const actions: Action[] = [];
+
+        if (this.isEditMode && matchesKeystroke(event, 'Escape')) {
+            this.isEditMode = false;
+        }
+
+        if (this.getSelectedElements(element.root).length > 0) {
+            if (!this.isEditMode && matchesKeystroke(event, 'KeyR', 'alt')) {
```
martin-fleck-at, line 85: Is it on purpose that we are only allowed to switch into edit mode if there are selected elements but we are allowed to get out of the edit mode without selection? If so, what is the reasoning?

packages/client/src/features/change-bounds/resize-tool.ts
```diff
@@ -0,0 +1,157 @@
...
+    protected isValidSize(element: SModelElement & BoundsAware, size: Dimension): boolean {
+        return isValidSize(element, size);
+    }
+
+    protected isValidMove(element: SModelElement & BoundsAware, newPosition: Point): boolean {
+        return isValidMove(element, newPosition, this.tool.movementRestrictor);
+    }
+
+    protected getSelectedElements(root: SModelRoot): (SModelElement & BoundsAware & Selectable)[] {
+        return Array.from(
```
martin-fleck-at, line 149: minor: If you were to inject the selection service in the tool, you could make this method a bit easier to read:
```
    protected getSelectedElements(): SelectableBoundsAware[] {
        const index = this.tool.selectionService.getModelRoot().index;
        const selectedElements = this.tool.selectionService.getSelectedElementIDs();
        return getElements(index, selectedElements, isSelectableAndBoundsAware);
    }
```

martin-fleck-at (changes requested):
Thank you very much @aylin-sarioglu!

I had a look at the change and everything seems to work well. However, there is a problem with the keybindings in the Theia integration and I'd love to have the code a bit more customizable and moved into a separate directory (please see my comments inline).

packages/client/src/features/change-bounds/resize-tool.ts
aylin-sarioglu: It was on purpose. Otherwise, the user would need to select an element again to cancel the edit mode if they have lost the selection somehow.

packages/client/src/features/change-bounds/resize-tool.ts
```diff
@@ -0,0 +1,157 @@
...
+    override keyDown(element: SModelElement, event: KeyboardEvent): Action[] {
+        const actions: Action[] = [];
+
+        if (this.isEditMode && matchesKeystroke(event, 'Escape')) {
+            this.isEditMode = false;
+        }
+
+        if (this.getSelectedElements(element.root).length > 0) {
+            if (!this.isEditMode && matchesKeystroke(event, 'KeyR', 'alt')) {
+                this.isEditMode = true;
```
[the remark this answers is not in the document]
aylin-sarioglu: Yes, user notifications will be provided in one of the upcoming PR. A message, either if the mode is on or off, will be displayed and directly called from resize-key-tool, so a cursor change is unnecessary as the notification is clear enough.

packages/client/src/features/change-bounds/resize-tool.ts
```diff
@@ -0,0 +1,157 @@
...
+            }
+
+            if (this.isEditMode) {
+                const selectedElements = this.getSelectedElements(element.root);
+                if (event.key === '+') {
+                    for (const elem of selectedElements) {
```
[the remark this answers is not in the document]
aylin-sarioglu: >  I'm also wondering whether we should use a level of indirection through a dedicated command here so we do not force the server resize.

Could you explain this in more detail?

packages/client/src/features/change-bounds/resize-tool.ts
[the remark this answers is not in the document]
aylin-sarioglu: >  Unfortunately Alt+R is already used in the Theia integration for Reveal in Explorer which is why this keybinding will not get triggered properly. I think we need to find a good alternative for it.

1. One possibility would be to prevent the event from coming to Theia. Because especially when the element is currently focused/selected, only the associated actions should be available.
2. Replace ALT+R with e.g. one of the function keys (F1...F12).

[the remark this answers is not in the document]
martin-fleck-at: Just to follow up on this: the function keys are also quite risky in a browser behavior as at least F1 (Help) and F5 (Reload). So we definitely need to find a valid keybinding before we can merge this.

packages/client/src/features/tools/resize-key-tool.ts
```diff
@@ -0,0 +1,191 @@
...
+
+        return undefined;
+    }
+
+    protected minSize(element: SelectableBoundsAware): Action | undefined {
+        const { x, y } = element.bounds;
+        const width = minWidth(element);
+        const height = minHeight(element);
+
+        console.log(width, height);
```
martin-fleck-at, line 144: Log needs to be removed.

martin-fleck-at (changes requested):

packages/client/src/features/tools/di.config.ts
ndoschek, line 1: As mentioned in [a PR comment for the move/zoom feature](https://github.com/eclipse-glsp/glsp-client/pull/241#discussion_r1189611152), this module could then also be part of the accessbility composite module.

packages/client/src/features/tools/resize-key-tool.ts
```diff
@@ -0,0 +1,189 @@
...
+        return matchesKeystroke(event, 'Escape');
+    }
+
+    protected matchesMinSizeKeystroke(event: KeyboardEvent): boolean {
+        return matchesKeystroke(event, 'Digit0', 'ctrl');
+    }
+
+    protected matchesReduceSizeKeystroke(event: KeyboardEvent): boolean {
+        return matchesKeystroke(event, 'Minus');
+    }
```
ndoschek, line 188: As mentioned in the move/zoom PR, it would be great to support numpad keys as well: https://github.com/eclipse-glsp/glsp-client/pull/241#discussion_r1189568526

ndoschek (commented):
Thanks @aylin-sarioglu for this contribution! :tada: 

In general it looks good to me, I added some comments inline, which mainly refer to comments I added in the other two open PRs.

I agree with @martin-fleck-at here, that we need to find a suitable shortcut for this feature in the theia-integration.
Also, if we can go with the same approach as I suggested in the move/zoom PR to decouple the functionality from key listener and extract the functionality into a separate Action, that would be even better.

aylin-sarioglu:
@martin-fleck-at @ndoschek 

currently the eslint configuration is invalid and causes an error. The configuration needs to be changed. See the following links:

https://github.com/typescript-eslint/tslint-to-eslint-config/issues/856

https://stackoverflow.com/questions/63961803/eslint-says-all-enums-in-typescript-app-are-already-declared-in-the-upper-scope

tortmayr:
> @martin-fleck-at @ndoschek
> 
> currently the eslint configuration is invalid and causes an error. The configuration needs to be changed. See the following links:
> 
> [typescript-eslint/tslint-to-eslint-config#856](https://github.com/typescript-eslint/tslint-to-eslint-config/issues/856)
> 
> https://stackoverflow.com/questions/63961803/eslint-says-all-enums-in-typescript-app-are-already-declared-in-the-upper-scope

@aylin-sarioglu Good catch. We will include this change the next time we update the glsp config packages. For now I recommend to simple suppress the warning. We already do this in [a couple of other places](https://github.com/eclipse-glsp/glsp-client/blob/d3b810d2d295b12ef4a2854751ba55a309aad21a/packages/protocol/src/action-protocol/element-text-editing.ts#L167).

ndoschek (changes requested):
Hi @aylin-sarioglu, thank you for the updates! :+1: 

Overall it looks good to me.
However, I found a small issue regarding the reset to the default view size.
It seems that after multiple resize actions, the reset does not work properly, after only one it works as expected, as far as I can tell. Also if I press `Ctrl+0` a second time, the default size is restored.
I recorded a short clip of this:

https://github.com/eclipse-glsp/glsp-client/assets/47414588/4cdaa105-ff17-4c73-9983-6bc03498fd65

Could you please have a look? Thanks!


And as you might have already read in the other PRs, I'd suggest to group the feature in the accessibility feature directory like so:
image

Please also squash your commits and add the key points of your changes as commit message, thanks!

aylin-sarioglu:
> Hi @aylin-sarioglu, thank you for the updates! +1
> 
> Overall it looks good to me. However, I found a small issue regarding the reset to the default view size. It seems that after multiple resize actions, the reset does not work properly, after only one it works as expected, as far as I can tell. Also if I press `Ctrl+0` a second time, the default size is restored. I recorded a short clip of this:
> 
>  glsp-accessibility-resize.mp4 
> Could you please have a look? Thanks!
> 
> And as you might have already read in the other PRs, I'd suggest to group the feature in the accessibility feature directory like so: image
> 
> Please also squash your commits and add the key points of your changes as commit message, thanks!

Thanks @ndoschek, the issue is reported and can be found here [Issue#1031](https://github.com/eclipse-glsp/glsp/issues/1031)

ndoschek (approved):
Thank you @aylin-sarioglu, changes look good to me and thanks for the quick update!
I merged the second PR, so we have again a few conflicts. Could you have a look and resolve them? Thanks!

Just a small remark, perhaps you should base the different feature branches on top of each other for the upcoming PRs to save you some time there.

aylin-sarioglu:
> Thank you @aylin-sarioglu, changes look good to me and thanks for the quick update! I merged the second PR, so we have again a few conflicts. Could you have a look and resolve them? Thanks!
> 
> Just a small remark, perhaps you should base the different feature branches on top of each other for the upcoming PRs to save you some time there.

Hello @ndoschek , I needed to add some changes, as after merging, both the zoom and resize tools overlapped in their functionality, as they both need to use '+' & '-'. It works now, you may want to look at the changes in [resize-key-tools.ts](https://github.com/eclipse-glsp/glsp-client/blob/807bfd81e3c24bd3a82000c4104e5606e6ddfef2/packages/client/src/features/accessibility/resize-key-tool/resize-key-tool.ts).

ndoschek (approved):
Thank you @aylin-sarioglu!
Changes look great to me! :sparkles: 

I tested now all three accessibility features together and it works great!

Also, I'm looking forward to the user notifications (which you added in the PR description) will be contributed in a follow-up PR.

Thanks again! :tada:

PR #241, opened 2023-04-19
Add keyboard support for moving and zooming elements and viewport

aylin-sarioglu:
This PR is part of my diploma research conducted at the [Model Engineering lab](https://me.big.tuwien.ac.at/) of TU Wien in the [Business Informatics Group](https://www.big.tuwien.ac.at/). It introduces the **move** and **zoom** functionality.

### Move elements
1. Select one or multiple elements
2. Use the arrow keys to move the element(s) accordingly

### Move viewport
1. Focus is on viewport , no elements are selected
2. Use the arrow keys to move the viewport accordingly
 
### Zoom  elements
1. Select one or multiple elements
2. Use `+` or `-` to zoom in/out  into the direction of the element or elements
3. Use `CTRL + 0` to reset to default zoom level

### Zoom  viewport
1. Focus is on viewport , no elements are selected
2. Use `+` or `-` to zoom in/out in viewport
3. Use `CTRL + 0` to reset to default zoom level

@martin-fleck-at, @ndoschek

Part of https://github.com/eclipse-glsp/glsp/issues/1029

PR #242, opened 2023-04-19
Add keyboard support for searching elements

aylin-sarioglu:
This PR is part of my diploma research conducted at the [Model Engineering lab](https://me.big.tuwien.ac.at/) of TU Wien in the [Business Informatics Group](https://www.big.tuwien.ac.at/). It introduces the **search** functionality.

### Search 
1. Focus is on viewport or element(s)
2. Use the shortcut  `CTRL + F` to display search field
3. Type in search term (label or type of elements)
4. Elements which fulfill the condition will be outlined, elements which **do not** fulfill the condition will become transparent.
5. Use arrow up or down to iterate through the returned list --> will jump to corresponding element and outline it

@martin-fleck-at, @ndoschek

### DIFF
--- a/examples/workflow-glsp/src/di.config.ts
+++ b/examples/workflow-glsp/src/di.config.ts
@@ -22,6 +22,7 @@ import {
     DiamondNodeView,
     editLabelFeature,
     GEdge,
+    glspAccessibilityModule,
     GLSPGraph,
     GLSPProjectionView,
     GridSnapper,
@@ -84,7 +85,7 @@ export function initializeWorkflowDiagramContainer(
     widgetId: string,
     ...containerConfiguration: ContainerConfiguration
 ): Container {
-    initializeDiagramContainer(container, workflowDiagramModule, directTaskEditor, ...containerConfiguration);
+    initializeDiagramContainer(container, workflowDiagramModule, directTaskEditor, glspAccessibilityModule, ...containerConfiguration);
     overrideViewerOptions(container, {
         baseDiv: widgetId,
         hiddenDiv: widgetId + '_hidden'
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   },
   "devDependencies": {
     "@eclipse-glsp/dev": "1.1.0-next.164cf99.124",
+    "@types/lodash": "4.14.191",
     "@types/node": "16.x"
   },
   "engines": {
--- a/packages/client/css/autocomplete-palette.css
+++ b/packages/client/css/autocomplete-palette.css
@@ -1,0 +1,22 @@
+/********************************************************************************
+ * Copyright (c) 2023 Business Informatics Group (TU Wien) and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+.autocomplete-palette {
+    position: absolute;
+    left: 20px;
+    top: 20px;
+    width: 400px;
+}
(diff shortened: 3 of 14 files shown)


ndoschek:
Hi @aylin-sarioglu, thank you very much for your contribution. I'll look at it at the beginning of next week.
Could you please rebase to latest master in the meantime to resolve the conflicts? Thank you :)

ndoschek (changes requested):
Thank you very much for this great contribution @aylin-sarioglu! :+1: 

Looks good to me overall, I tested mainly with the standalone example and added a comment below regarding the theia-integration.

I added some comments inline and I came across two issues, which I could not directly link to the code, would be great if you could have a look:
- If I completely deselect all elements via Escape, the search widget still holds the last searched element information. Is this the expected behaviour? For example, I search for the `Brew` task, then deselect all elements, select the `ChkWater` element and open the search widget, the `Brew` task is again selected and centered. I'm not sure if this is what I would expect.
- On startup, if no search was done and no element is selected: Is there a reason why the first element is selected and centered if the search command palette is opened? I find that a bit confusing tbh.

**theia-integration**:
Also, I'm sure you are aware of this, but for the sake of completeness: The functionality does not yet work in the theia-integration. As mentioned in [the PR comment for the move/zoom feature](https://github.com/eclipse-glsp/glsp-client/pull/241#pullrequestreview-1420035581), we have the benefit in this case, that the key listener is already directly triggering an action, which is great!
In this case, we can then make use of a standard 'find in editor' command (i.e. `CommonCommands.FIND`) and apply the same pattern as we do for the Undo/Redo commands.

This will work easily as:
in `packages/theia-integration/src/browser/diagram/glsp-diagram-keybinding.ts`:
```typescript
registerKeybindings(registry: KeybindingRegistry): void {
        [
           ....
           {
                command: CommonCommands.FIND.id,
                context: this.diagramKeybindingContext.id,
                keybinding: 'ctrlcmd+f'
            }
            ....
```
and in `packages/theia-integration/src/browser/diagram/glsp-diagram-commands.ts`:
```typescript
registerCommands(registry: CommandRegistry): void {
   ...
   registry.registerHandler(
            CommonCommands.FIND.id,
            new GLSPCommandHandler(this.shell, {
                actions: () =>
                    SetUIExtensionVisibilityAction.create({
                        extensionId: 'search-autocomplete-palette',
                        visible: true
                    }),
                alwaysVisible: true
            })
        );
}
```

Thanks again for this cool new accessiblity feature! :tada:

ndoschek (commented):
Thank you @aylin-sarioglu for the updates!
Overall it looks good, I just have three comments left:

1. > On startup, if no search was done and no element is selected: Is there a reason why the first element is selected and centered if the search command palette is opened? I find that a bit confusing tbh.
I [asked this earlier](https://github.com/eclipse-glsp/glsp-client/pull/242#pullrequestreview-1420170997), is there a specific reason? Couldn't we just deselect all and center the diagram for example?

2. Also, now if one element was found, and I open a new search, the selection of the first searched element is kept, and all following search results are added to the selection. Should we clear the selections on each search, what do you think? I don't have no hard opinion, but also I'm not sure if adding them to the selection is what I would expect.

As mentioned in the previous PR, I think it would be good to group all the accessiblity features into on directory, I'd suggest this file structure:
image

aylin-sarioglu:
> Thank you @aylin-sarioglu for the updates! Overall it looks good, I just have three comments left:
> 
> 1. > On startup, if no search was done and no element is selected: Is there a reason why the first element is selected and centered if the search command palette is opened? I find that a bit confusing tbh.
>    > I [asked this earlier](https://github.com/eclipse-glsp/glsp-client/pull/242#pullrequestreview-1420170997), is there a specific reason? Couldn't we just deselect all and center the diagram for example?
> 2. Also, now if one element was found, and I open a new search, the selection of the first searched element is kept, and all following search results are added to the selection. Should we clear the selections on each search, what do you think? I don't have no hard opinion, but also I'm not sure if adding them to the selection is what I would expect.
> 
> As mentioned in the previous PR, I think it would be good to group all the accessiblity features into on directory, I'd suggest this file structure: image
> 
> Please also squash your commits and add the key points of your changes as commit message, thanks!

Thank you  @ndoschek for your feedback. 

Regarding to your comments: 

1. Unfortunately, I cannot change the behavior as we are using an old version of the autocompleter library ([auto-complete-widget.ts](https://github.com/eclipse-glsp/glsp-client/blob/master/packages/client/src/base/auto-complete/auto-complete-widget.ts#L52)). The version we have is outdated (4 years) and the current version only supports this functionality. Changing it would probably cause side effects in this PR. Maybe a future PR could change this behaviour where the library is also updated.

Comment 2. and the folder structure are fixed now.

ndoschek (approved):
Thank you for the quick updates @aylin-sarioglu! :+1: 
Changes look good to me now :+1:

PR #243, opened 2023-04-20
Add a reason for requesting and setting markers

planger:
Also track feedback by reason to replace only those on `SetMarkersAction` that have the same reason.

https://github.com/eclipse-glsp/glsp/issues/980

For review, please see [the comment on which changes belong together](https://github.com/eclipse-glsp/glsp/issues/980#issuecomment-1516867263).


tortmayr (changes requested):
Thanks Philip, support for live validation is definitely a nice feature addition.
I did not have time to test the changes yet and only did a preliminary code review for now.

planger:
Thank you for the excellent and quick review! I've applied your suggestions in 83e98974abba81fd76f6f50a0add5b2c9f1484e3

tortmayr (changes requested):
The changes look good to me but looks like the tests are failing.
I guess the tests for `RequestMarkesActions`|`SetMarkersAction` in the protocol package need to be updated

tortmayr (approved):
LGTM! 👍🏼

PR #245, opened 2023-04-24
GLSP-975 Move glsp-server types into protocol packages

tortmayr:
Ensures that the `@eclipse-glsp/protocol` package also provides the glsp server interfaces. This allows type reusage in usecases where the GLSP server is not running in an external process and is instead accessed directly.

Part of https://github.com/eclipse-glsp/glsp/issues/975


sgraband (approved):
Thank you for the changes @tortmayr! LGTM :+1:

PR #247, opened 2023-05-19
GLSP-975: Implement NodeGlspClient

tortmayr:
In addition, to the classic json-rpc client the protocol package now also offers a `GLSPClient` that communicates directly with a `GLSPServer` without any communication layer between. This client can be used in cases where the (Node) server and client are running in the same context i.e. process.
- Part of https://github.com/eclipse-glsp/glsp/issues/975

In addition, this PR contains the following changes:
- Refactor `onActionMessage` method of `GLSPClient` to return a disposable for deregistration
- Refactor  connection state checking in `BaseJsonrpcClient` 
- Add test classes for both GLSPClient implementations
- Update tsconfig to ensure that spec files are not part of the published package
- Part of https://github.com/eclipse-glsp/glsp/issues/1004


packages/protocol/src/client-server-protocol/jsonrpc/base-jsonrpc-glsp-client.ts
```diff
@@ -40,49 +35,41 @@ export class BaseJsonrpcGLSPClient implements GLSPClient {
...
         }
-        return true;
+        return this.resolvedConnection!;
+    }
+
+    /**
+     * @Deprecated use {@link checkConnection} instead
+     */
+    protected checkConnectionState(): boolean {
+        return this.checkedConnection !== undefined;
```
martin-fleck-at, line 72: Since this is only for backwards compatibility I think having this method is fine. In the long run it would be great to have one method that throws an exception (checkedConnection) and one that does a safe check, i.e., always return a boolean. A mixture of boolean and exception is unexpected.

packages/protocol/src/utils/test-util.spec.ts
```diff
@@ -0,0 +1,45 @@
+/********************************************************************************
```
martin-fleck-at, line 1: I think it is a bit unexpected to have a `test-util.spec.ts` without a `test-util.ts`. Could you explain what this file tests? If it is only the `MaybePromise` I believe we should add it to the `type-util.spec.ts`, what do you think?

packages/protocol/src/client-server-protocol/jsonrpc/base-jsonrpc-glsp-client.ts
```diff
@@ -154,7 +141,7 @@ export class BaseJsonrpcGLSPClient implements GLSPClient {
         this.state = ClientState.ServerError;
     }
 
-    protected isConnectionActive(): boolean {
+    public isConnectionActive(): boolean {
```
martin-fleck-at, line 144: unnecessary `public`

packages/protocol/src/client-server-protocol/jsonrpc/base-jsonrpc-glsp-client.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+    onClose = this.mockEvent;
+    listen(): void {}
+    onUnhandledNotification = this.mockEvent;
+    end(): void {}
+    onDispose = this.mockEvent;
+    dispose(): void {}
+    inspect(): void {}
+}
+
+describe('Base JSON-RPC GLSP Client', () => {
```
martin-fleck-at, line 69: Very nice test cases for the base GLSP client implementation! If I could only have one request, it would be that it also tests the `onActionMessage` method since this is part of the public API. Not sure how easy it is with the mock implementation though.

martin-fleck-at (changes requested):
Fantastic work @tortmayr! I really really like the generic base classes and the addition of test cases which will strengthen the trust in the implementations even more!

I only have a few comments and questions but other than that this PR is ready to be merged.

packages/protocol/src/client-server-protocol/jsonrpc/base-jsonrpc-glsp-client.ts
tortmayr: Yeah, we already have this two methods with
`checkedConnection` and `isConnectionActive`. 
As you already mentioned the `checkConnectionState()` method is just for backwards compatibility. But I would also fine with dropping this method entirely.

packages/protocol/src/utils/test-util.spec.ts
tortmayr: This modules provides utility methods that are only used in the test context (`expectToThrowAsync`). It imports from `chai` which is a test dependency => it should not be part of the production build. That`s is the reason why I used the `spec` suffix. 
While I agree that the `spec` suffix is a bit unexpected I don't think moving it into the `type-util.spec.ts` is a good idea.
I would rather drop the spec suffix and exclude the file explicitly from the tsconfig.

packages/protocol/src/client-server-protocol/jsonrpc/base-jsonrpc-glsp-client.spec.ts
tortmayr: We could add a test to verify that the corresponding method of the underlying message connection is triggered (`connection.onNotification(JsonrpcGLSPClient.ActionMessageNotification, handler)`) but other than that i don't think additional test cases makes sense because we would either test a artificial mock implementation that does not have anything in common with the actual handling or we would test the functionality provided by vscode-jsonrpc.

packages/protocol/src/utils/test-util.spec.ts
martin-fleck-at: Sounds good and works for me, thank you for the fast response!

packages/protocol/src/client-server-protocol/jsonrpc/base-jsonrpc-glsp-client.ts
martin-fleck-at: Since you marked it as deprecated, it is fine for me. I just wanted to note down my thoughts because it is indeed an unexpected method. Thank you for the clarification.

martin-fleck-at (approved):
LGTM!

PR #248, opened 2023-05-21
GLSP-813 Use Node example server instead of java variant

tortmayr:
Remove download script for Java Workflow Server and consume node variant via npm instead. Update README to reflect the changes.

Part of https://github.com/eclipse-glsp/glsp/issues/813


martin-fleck-at (approved):
Very nice! That will make testing features and changes much easier, thank you Tobias!

PR #249, opened 2023-05-22
GLSP-975: Rename `NodeGLSPClient`

tortmayr:
The `NodeGLSPClient` can also be used in a browser context => rename to `BaseGLSPClient`. This way we can introduce a consistent naming scheme for generic base client implementations: `Base<communication-channel>GLSPClient`. 
For instance if are going to provide a REST implementation in the future it could be called `BaseRestGLSPClient`.


martin-fleck-at (approved):
LGTM!

PR #255, opened 2023-06-29
Be more flexible when it comes to calculating the route of an edge

martin-fleck-at:
- Allow tolerance when comparing coordinates to detect duplicates
- Only apply duplicate detection on pure routing point kinds

Fixes https://github.com/eclipse-glsp/glsp/issues/976


tortmayr (approved):
LGTM! Thanks

PR #257, opened 2023-07-03
GLSP-979: Introduce 'deselectAll' flag for SelectAction

martin-fleck-at:
- Introduce optional flag that indicates deselection of all elements
- Introduce convenience creation methods for SelectAll action

https://github.com/eclipse-glsp/glsp/issues/979

### DIFF
--- a/packages/client/src/features/select/select-feedback-action.ts
+++ b/packages/client/src/features/select/select-feedback-action.ts
@@ -40,22 +40,19 @@ export namespace SelectFeedbackAction {
         return { ...SelectAction.create(options), kind: KIND };
     }
 
-    export function addSelection(options: { selectedElementsIDs: string[] }): SelectFeedbackAction {
-        return { ...SelectAction.addSelection(options), kind: KIND };
+    export function addSelection(selectedElementsIDs: string[]): SelectFeedbackAction {
+        return { ...SelectAction.addSelection(selectedElementsIDs), kind: KIND };
     }
 
-    export function removeSelection(options: { deselectedElementsIDs: string[] }): SelectFeedbackAction {
-        return { ...SelectAction.removeSelection(options), kind: KIND };
+    export function removeSelection(deselectedElementsIDs: string[]): SelectFeedbackAction {
+        return { ...SelectAction.removeSelection(deselectedElementsIDs), kind: KIND };
     }
 
-    export function setSelection(options: { selectedElementsIDs: string[] }): SelectFeedbackAction {
-        return { ...SelectAction.setSelection(options), kind: KIND };
-    }
-
-    export function deselectAll(options: object = {}): SelectFeedbackAction {
-        return { ...SelectAction.deselectAll(options), kind: KIND };
+    export function setSelection(selectedElementsIDs: string[]): SelectFeedbackAction {
+        return { ...SelectAction.setSelection(selectedElementsIDs), kind: KIND };
     }
 }
+
 @injectable()
 export class SelectFeedbackCommand extends Command {
     static readonly KIND = SelectFeedbackAction.KIND;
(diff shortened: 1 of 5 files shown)


packages/client/src/features/select/select-feedback-action.ts
```diff
@@ -36,9 +36,25 @@ export namespace SelectFeedbackAction {
         return Action.hasKind(object, KIND) && hasArrayProp(object, 'selectedElementsIDs') && hasArrayProp(object, 'deselectedElementsIDs');
     }
 
-    export function create(options?: { selectedElementsIDs?: string[]; deselectedElementsIDs?: string[] }): SelectFeedbackAction {
+    export function create(options?: { selectedElementsIDs?: string[]; deselectedElementsIDs?: string[] | boolean }): SelectFeedbackAction {
         return { ...SelectAction.create(options), kind: KIND };
     }
+
+    export function addSelection(options: { selectedElementsIDs: string[] }): SelectFeedbackAction {
```
tortmayr, line 43: I don't think that we need an options object here. We only use it in the create function because it's the default pattern to declare optional properties in an options object. Here we can just use plain arguments.

packages/protocol/src/action-protocol/element-selection.ts
```diff
@@ -45,14 +51,32 @@ export namespace SelectAction {
         return Action.hasKind(object, KIND) && hasArrayProp(object, 'selectedElementsIDs') && hasArrayProp(object, 'deselectedElementsIDs');
     }
 
-    export function create(options: { selectedElementsIDs?: string[]; deselectedElementsIDs?: string[] } = {}): SelectAction {
+    export function create(options: { selectedElementsIDs?: string[]; deselectedElementsIDs?: string[] | boolean } = {}): SelectAction {
+        const { deselectedElementsIDs, ...remainingOptions } = options;
         return {
             kind: KIND,
-            selectedElementsIDs: [],
```
tortmayr, line 51: I don't think that the spreading into remaining options is necessary here. We already use `options.selectedElementIds` and `options.deselectedElementIds` so there are no more remaining options 😉

packages/protocol/src/action-protocol/element-selection.ts
```diff
@@ -45,14 +51,32 @@ export namespace SelectAction {
...
+
+    export function removeSelection(options: { deselectedElementsIDs: string[] }): SelectAction {
+        return create(options);
+    }
+
+    export function setSelection(options: { selectedElementsIDs: string[] }): SelectAction {
+        return create({ ...options, deselectedElementsIDs: true });
+    }
+
+    export function deselectAll(options: object = {}): SelectAction {
```
tortmayr, line 77: Is this really needed. Isn't this allready handled by  `SelectAllAction.create(false)` ?

tortmayr (changes requested):
Thanks Martin. I really like the addition of the new select utility functions. They are quite handy!
I have a few minor inline comments

packages/client/src/features/select/select-feedback-action.ts
martin-fleck-at: I see, I thought it was to support spreading of additional options so I wanted to keep the pattern. But now I see what you mean and I changed it accordingly.

packages/protocol/src/action-protocol/element-selection.ts
martin-fleck-at: Good point! I didn't think of the `SelectAll`. I removed this method from here and the feedback selection.

martin-fleck-at:
@tortmayr Thank you for the quick feedback! All good points and I made the changes as you proposed.

tortmayr (approved):
LGTM!

PR #258, opened 2023-07-05
GLSP-1048: Keep focus if it was taken deliberately through an action

martin-fleck-at:
- Fix minor issue with default button not being present before init

Fixes https://github.com/eclipse-glsp/glsp/issues/1048


tortmayr (approved):
Good catch!
Change looks good to me and focus switching now works as expected.

PR #261, opened 2023-07-09
GLSP-1050 Introduce Event API

tortmayr:
- Introduce/reuse Event API from `vscode-jsonrpc`
- Rework notifier services to use new API
- Remove manual registration methods from notifier services

- Adapt test cases for `SelectionService` and remove test cases that were testing generic listener functionality that  is already provided (and therefore assumed tested) by `vscode-jsonrpc` Also: 
- Align prefix of injectable interfaces to consistently use the `I` prefix

Also:
- Update test launch configs to use an infinite timeout for easier debugging.
- Fix missing binding to `ISmodelRootListener` for the SelectionService.
- Fix wrong order in implementing edit mode listeners (oldValue,newValue => newValue, oldValue)
- Fix typo in `ToolPallete` (`lastActivebutton` => `lastActiveButton`)


packages/client/src/base/command-stack.ts
```diff
@@ -13,22 +13,38 @@
...
-import { CommandStack, ICommand, SModelRoot, SetModelCommand, TYPES, UpdateModelCommand, distinctAdd, remove } from '~glsp-sprotty';
+import { injectable, multiInject, optional, preDestroy } from 'inversify';
+import { CommandStack, Disposable, Emitter, Event, ICommand, SModelRoot, SetModelCommand, TYPES, UpdateModelCommand } from '~glsp-sprotty';
 
 /**
  * A hook to listen for model root changes. Will be called after a server update
  * has been processed
  */
-export interface SModelRootListener {
+export interface ISModelRootListener {
```
martin-fleck-at, line 23: I'm just slightly wondering whether we should consider a name like `IGModelRootListener` for this since we are already breaking API and https://github.com/eclipse-glsp/glsp/issues/865 is something that will also be done at some point. It is really the only one of the types that should be affected by this.

packages/client/src/base/selection-service.ts
```diff
@@ -53,14 +60,26 @@ export class SelectionService implements SModelRootListener {
...
+        this.selectionListeners.forEach(listener =>
+            this.onSelectionChanged(change => listener.selectionChanged(change.root, change.selectedElements, change.deselectedElements))
+        );
+    }
 
-    register(selectionListener: SelectionListener): void {
-        distinctAdd(this.selectionListeners, selectionListener);
+    @preDestroy()
+    dispose(): void {
+        this.onSelectionChangedEmitter.dispose();
```
martin-fleck-at, line 76: I'm a big fan of the "something implements Disposable so we have a toDispose collection"-pattern. But that is just a preference so we can keep it as is but I always appreciated that in the framework that I can add the toDipose stuff already when I override initialize somehow.

packages/client/src/features/tool-palette/tool-palette.ts
```diff
@@ -99,7 +99,7 @@ export class ToolPalette extends AbstractUIExtension implements IActionHandler,
     protected initializeContents(_containerElement: HTMLElement): void {
         this.createHeader();
         this.createBody();
-        this.lastActivebutton = this.defaultToolsButton;
+        this.lastActiveButton = this.defaultToolsButton;
```
martin-fleck-at, line 102: Love it!

packages/protocol/src/utils/event.ts
```diff
@@ -0,0 +1,92 @@
...
+ */
+export interface Event<T> extends jsonrpc.Event<T> {
+    /**
+     *
+     * @param listener The listener function will be called when the event happens.
+     * @param thisArgs The 'this' which will be used when calling the event listener.
+     * @param disposables An array to which the {@link Disposable} for removing the listener will be added.
+     * @returns a {@link Disposable} to remove the listener again.
+     */
+    (listener: (e: T) => any, thisArgs?: any, disposables?: Disposable[]): Disposable;
```
martin-fleck-at, line 46: So I assume we override this because we ant to use our own "Disposable", right? I'm just wondering because at the moment our own disposable does not have any extra functionality. Also, should our Disposable now also extend `jsonrpc.Disposable`?

I think I'm just a bit confused as to why we need our own types that are basically the same (with the exception of exchanging `Function` with `SafeFunction` in the EmitterOptions.

packages/protocol/src/utils/event.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+/**
+ * Optional options that can be passed to the constructor
+ * of an {@link Emitter}.
+ */
+export interface EmitterOptions {
```
martin-fleck-at, line 52: Should that extends the `EmitterOptions` from jsonrpc? Cause otherwise the constructor might break at some point.

martin-fleck-at (changes requested):
Change looks good to me and works as expected, I just have a few comments that I would like to get clarified first before we merge this.

packages/client/src/base/command-stack.ts
tortmayr: Wouldn't i make more sense to this as part of https://github.com/eclipse-glsp/glsp/issues/865?
Seems strange to me to use `GModelRoot` right now since the type does not exist.

packages/protocol/src/utils/event.ts
tortmayr: We don't strictly need our own types this has been added to be more future proof and for developer convenience.
By overriding with our own types we ensure that things like auto import etc. use the correct types as well (from @eclipse-glsp/client and not vscode-jsonrpc).
In addition, it gives us the flexibility to add custom functionality to our types later on and also ensure that we could move away from vscode-jsonrpc if we decide at some point that we don't want to use it anymore.
Another indirect benefit is that we can add more documentation to the event classes 😉 

Ensuring compatibility with the vscode disposable by extending sounds like a good idea to me.

packages/protocol/src/utils/event.ts
tortmayr: Unfortunatley the `EmitterOptions` are not part of the main export of `vscode-jsonrpc`.

packages/protocol/src/utils/event.ts
martin-fleck-at: Thank you for the clarification! I expected as much but I just wanted to make sure I fully understand the intention behind it. With that, this looks totally fine to me.

packages/protocol/src/utils/event.ts
martin-fleck-at: Got it!

martin-fleck-at (approved):
LGTM!

PR #262, opened 2023-07-11
Use feedback dispatcher for change bounds feedback

planger:
Fixes https://github.com/eclipse-glsp/glsp/issues/941

How I tested this:
1. move and resize elements and then hit Escape to see if the feedback is properly reset
2. install a timer to periodically send an updatemodelaction from the server and see if the visual feedback for move isn't disturbed and correctly re-applied after each model update from the server


martin-fleck-at (approved):
Works exactly as advertised, thank you for the great fix! LGTM!

PR #265, opened 2023-07-12
Add real progress reporting in favor of timeout

planger:
* Remove `ServerMessageAction.timeout`
* Add actions for reporting progress

https://github.com/eclipse-glsp/glsp/issues/1019


packages/protocol/src/action-protocol/client-notification.ts
```diff
@@ -114,3 +108,130 @@ export namespace ServerMessageAction {
...
+}
+
+export namespace EndProgressAction {
+    export const KIND = 'endProgress';
+
+    export function is(object: any): object is EndProgressAction {
+        return Action.hasKind(object, KIND) && hasStringProp(object, 'progressId');
+    }
+
+    export function create(
```
tortmayr, line 225: I don't know if we have other occurrences of this pattern but if the create function has one required and one optional parameter wrapping the optional parameter in an options object seems a bit overkill to me.

tortmayr (changes requested):
Change looks mostly good to me. I have a few minor nitpicks.
Will we are at it could please either open an issue to update the protocol on the 2.0 branch of the [website-source repo](https://github.com/eclipse-glsp/glsp-website-source/tree/2.0) or provide a PR directly?

planger:
@tortmayr Thanks for the feedback, I've addressed your comments and updated the other changes accordingly.
Also I've extended the description of https://github.com/eclipse-glsp/glsp/issues/1019 to list all required changes, including the one for the website.

tortmayr (approved):
LGTM! 👍🏼

PR #266, opened 2023-07-14
GLSP-1054: Rename di.config modules

tortmayr:
- Rename `di.config.ts` files to  `xyz-module.ts`
- Remove default export from di.config files
- Remove inconsistent prefixing of modules with `glsp`. All glsp modules are now unprefixed, if a module name conflicts with an existing sprotty module, the sprotty module is reexported with a `sprotty` prefix
- Aligned/renamed some module names

Also: 
- Move `focus-change-action` and `focus-tracker` to `focus` subdirectory

 Fixes https://github.com/eclipse-glsp/glsp/issues/1054


packages/client/src/glsp-sprotty/index.ts
```diff
@@ -342,3 +317,31 @@ import viewportModule from 'sprotty/lib/features/viewport/di.config';
...
+
+export {
+    buttonModule,
+    defaultModule,
+    edgeEditModule,
+    edgeIntersectionModule,
+    edgeLayoutModule,
+    expandModule,
+    fadeModule,
+    viewportModule as glspViewportModule,
```
planger, line 329:
```suggestion
    viewportModule as sprottyViewportModule,
```

Shouldn't this be prefixed with `sprotty`?

planger (commented):
Thank you! This is now imho much clearer to adopters.
Everything seems to work fine and code looks good.

However, I have one question inline above.

planger (approved):

PR #267, opened 2023-07-18
WIP: Introduce feature modules

tortmayr:
Introduce custom container modules that allow to specify required modules and enable conditional loading i.e. only load if the all required modules have been loaded before. Each feature module bind a `featureId`(symbol) when loaded which can be used by other feature modules to identify whether a required feature module is bound.

Prime usecase for this is to facilitate `extensionModules` i.e. modules that build on top of the functionality provided by a base module (and optionally customize,rebind certain services) for certain integration scenarios (standalone, theia, vscode).

This also reduces the customization overhead e.g.  application specific integrations as it's no longer necessary to manually customize the behavior if certain default modules should not be included. For instance if an adopter doesn't want to provide the `export` functionality he can simply remove the `exportModule` and any application specific customization (`standaloneExportModule`, `theiaExportModule`) will not be loaded in the container as well.

In addition, this change is also essential if we decide to switch to a `capability` based approach similar to LSP at some point. (i.e. the server defines it`s capabilities and the client dynamically loads the modules required for this capabilities).

I also have prepared a small demo branch to see how this proposal looks like in the Theia integration
https://github.com/eclipse-glsp/glsp-theia-integration/tree/example-usage


packages/protocol/src/utils/di-util.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+                        bind('Foo').toConstantValue('Foo');
+                    },
+                    { requires: moduleA }
+                );
+                container.load(moduleWithNoRequirements, moduleA);
+                expect(container.isBound(moduleA.featureId)).to.be.true;
+                expect(container.isBound(moduleWithNoRequirements.featureId)).to.be.false;
+                expect(container.isBound('Foo')).to.be.false;
+            });
+            it('Should not load a feature module with not loaded module and `failOnUnmetRequires`= false', () => {
```
planger, line 64:
```suggestion
            it('Should not load a feature module with missing required module and `failOnUnmetRequires`= false', () => {
```

packages/protocol/src/utils/di-util.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+                    },
+                    { requires: [moduleA, moduleB] }
+                );
+                container.load(moduleA, moduleWithUnmetRequirements);
+                expect(container.isBound(moduleA.featureId)).to.be.true;
+                expect(container.isBound(moduleWithUnmetRequirements.featureId)).to.be.false;
+                expect(container.isBound('Foo')).to.be.false;
+            });
+        });
+        it('Should not load a feature module with not loaded required modules and `failOnUnmetRequires`= false', () => {
```
planger, line 104:
```suggestion
        it('Should not load a feature module with missing required modules and `failOnUnmetRequires`= false', () => {
```

packages/protocol/src/utils/di-util.ts
```diff
@@ -81,6 +90,95 @@ export interface ModuleConfiguration {
...
+
+    constructor(registry: interfaces.ContainerModuleCallBack, options: FeatureModuleOptions = {}) {
+        super((bind, unbind, isBound, ...rest) => {
+            if (this.configure(bind, isBound)) {
+                registry(bind, unbind, isBound, ...rest);
+            }
+        });
+        this.featureId = options.featureId ?? this.createFeatureId();
+        this.requires = options.requires;
+        this.failOnUnmetRequires = options.failOnUnmetRequires ?? false;
```
planger, line 131: I'd be in favor of activating this by default as it otherwise makes debugging even harder, since now not only it will likely behave strangely at runtime, but also what has been explicitly bound in the adopters code will not be bound at all. At least we should always print an error to the console.

packages/client/src/features/viewport/viewport-modules.ts
```diff
@@ -30,17 +30,29 @@ import {
...
+ * Feature module that is intended for the standalone deployment of GLSP (i.e. plain webapp)
+ * When integrated into an application frame (e.g Theia/VS Code) this module is typically omitted and/or replaced
+ * with an application native module.
+ */
+export const standaloneViewportModule = new FeatureModule(
+    (bind, _unbind, isBound) => {
+        const context = { bind, isBound };
+        bindAsService(context, TYPES.KeyListener, CenterKeyboardListener);
+    },
+    { requires: viewportModule }
```
planger, line 57: A general thought: why don't we automatically bind the required module (`viewportModule`) if it hasn't been bound already? Maybe this adds too much magic, but on the other hand that could be also what someone may expect when they add a `requires` somewhere.

planger (commented):
Excellent work, thank you very much! I'm sure this will avoid some painful debugging in the future and also gives a more systematic way of managing modules. Just a few minor thoughts / suggestions inline.

packages/client/src/features/viewport/viewport-modules.ts
tortmayr: Mhh, this would break the current intended behavior for default modules + application specific modules.
If a default module is missing we have to assume that the adopter intentionally removed it during the container initialization.
In that case we simply don't want to load and modules that require/extend behavior of the removed default module.
Loading the required module in this would magically revert the adopters intentional removal.

The main intended use case (at least from my point of view) is conditional module loading.

For instance if a user removes the `copyPasteModule` because it should not be supported in their Diagram Editor than we also don't want to load any c&p related addtions like context menu entries or keyboard short cuts.
This is also the reason why `failsOnMissingRequired` is false per default.

Maybe we could rework the API a little bit? e.g:
```ts
/**
 * Optional constructor options for {@link FeatureModule}s.
 */
export interface FeatureModuleOptions {
   loadIfPresent?: MaybeArray<FeatureModule>; // set of module that need to be loaded as precondition for this module to load. Does not throw an error or add any of the prerequistes
   requires: MaybeArray<FeatureModule> // set of hard required modules. Will throw an error if the requirements are not ment.
    featureId?: symbol;
}
`
Wdyt?

planger: Yes, I see what you mean. I also would like to avoid making this more complex than it needs to be.
From an adopter's point of view, I see two values in this system:
1. Avoiding to load dependents when a module isn't loaded or unloaded.
2. Throw an error if a user tries to load a module for which the dependencies aren't available.

Would you agree? Is there something I miss?
If it is exactly those two, we may want to optimize for them, because currently we don't support both of them at the same time (because `failsOnMissingRequired` is either `true` or `false`), whereas we want to just unload for use case 1 and fail for use case 2. And I'm wondering when `failsOnMissingRequired` would ever be very useful as soon as someone unloads a module purposefully, if this module has dependencies (where they don't expect an error, but they will always retrieve one).
Would it help if we have different behavior for an explicit `unload` call (use case 1), where we don't throw any errors but just unload all dependents and an explicit load call where we throw an error if any of the explicitly mentioned modules have missing dependencies?

I'd also be fine if we first address use case 1 with your original proposal and don't put use case 2 in the focus at all. For some reason, I was mainly thinking about that use case as I read `failOnMissingRequired` and remember cases where adopters ran into issues, because they loaded a module without knowing that they need to load a dependency too.

tortmayr: >Yes, I see what you mean. I also would like to avoid making this more complex than it needs to be.
>From an adopter's point of view, I see two values in this system:

> 1. Avoiding to load dependents when a module isn't loaded or unloaded.
> 2. Throw an error if a user tries to load a module for which the dependencies aren't available.

Agreed, those are the values of the system.

>If it is exactly those two, we may want to optimize for them, because currently we don't support both of them at the same time (because failsOnMissingRequired is either true or false), whereas we want to just unload for use case 1 and fail for use case 2. And I'm wondering when failsOnMissingRequired would ever be very useful as soon as someone unloads a module purposefully, if this module has dependencies (where they don't expect an error, but they will always retrieve one).

I would argue that a mixed case isn't very useful anyways. What are the usecases that would require the module to not load for one dependent module and fail for another? And how would we resolve conflicting cases i.e. when dependencies are missing that require both not loading and failing?


We could also argue, that the `fail on required module` case is already handled decently enough by inversify. (missing binding error which indicates that a module is missing) and just cover the `not load on missing dependency usecase´ for now.

>I'd also be fine if we first address use case 1 with your original proposal and don't put use case 2 in the focus at all. For some reason, I was mainly thinking about that use case as I read failOnMissingRequired and remember cases where adopters ran into issues, because they loaded a module without knowing that they need to load a dependency too.

I see. Maybe we could also mitigate this issue by catching the error throw by inversify and rethrow/print a more clear error message?

planger: > I see. Maybe we could also mitigate this issue by catching the error throw by inversify and rethrow/print a more clear error message?

Yes that could possibly help! But we can also move this out of this PR imho (even after the Release if necessary). As you prefer and depending on how much additional effort that would case.

Thanks!

tortmayr:
@planger I'm going to merge this (as it is a preconditions for some of my pending PRs).
If you encounter any additional issues I'm going to address them in a follow-up PR.

PR #269, opened 2023-07-20
GLSP-77: Allow WebSocket connections to reconnect after interrupt

ndoschek:
- Introduce GLSPWebSocketProvider to allow websocket connections to reconnect after interrupt
- Make use of ws provider in standalone example

Part of https://github.com/eclipse-glsp/glsp/issues/77

---
Related to: https://github.com/eclipse-glsp/glsp-server-node/pull/54
Related to: https://github.com/eclipse-glsp/glsp-server/pull/208

Testing:
- To simulate interrupts, you can use the corresponding testing branch [`issues/77-testing`](https://github.com/eclipse-glsp/glsp-client/tree/issues/77-testing). This automatically interrupts the example twice with an timeout of 5s in between.

### DIFF
--- a/examples/workflow-standalone/src/app.ts
+++ b/examples/workflow-standalone/src/app.ts
@@ -18,15 +18,17 @@ import 'reflect-metadata';
 import {
     ApplicationIdProvider,
     BaseJsonrpcGLSPClient,
-    configureServerActions,
     EnableToolPaletteAction,
     GLSPActionDispatcher,
     GLSPClient,
     GLSPDiagramServer,
-    listen,
+    GLSPWebSocketProvider,
     RequestModelAction,
     RequestTypeHintsAction,
-    TYPES
+    ServerMessageAction,
+    ServerStatusAction,
+    TYPES,
+    configureServerActions
 } from '@eclipse-glsp/client';
 import { join, resolve } from 'path';
 import { MessageConnection } from 'vscode-jsonrpc';
@@ -34,20 +36,22 @@ import createContainer from './di.config';
 const port = 8081;
 const id = 'workflow';
 const diagramType = 'workflow-diagram';
-const websocket = new WebSocket(`ws://localhost:${port}/${id}`);
 
 const loc = window.location.pathname;
 const currentDir = loc.substring(0, loc.lastIndexOf('/'));
 const examplePath = resolve(join(currentDir, '../app/example1.wf'));
 const clientId = ApplicationIdProvider.get() + '_' + examplePath;
 
-const container = createContainer();
-const diagramServer = container.get<GLSPDiagramServer>(TYPES.ModelSource);
+const webSocketUrl = `ws://localhost:${port}/${id}`;
+
+let container = createContainer();
+let diagramServer = container.get<GLSPDiagramServer>(TYPES.ModelSource);
 diagramServer.clientId = clientId;
 
-listen(websocket, connection => initialize(connection));
+const wsProvider = new GLSPWebSocketProvider(webSocketUrl);
+wsProvider.listen({ onConnection: initialize, onReconnect: reconnect, logger: console });
 
-async function initialize(connectionProvider: MessageConnection): Promise<void> {
+async function initialize(connectionProvider: MessageConnection, isReconnecting = false): Promise<void> {
     const client = new BaseJsonrpcGLSPClient({ id, connectionProvider });
 
     await diagramServer.connect(client);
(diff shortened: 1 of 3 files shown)


packages/protocol/src/client-server-protocol/jsonrpc/ws-connection-provider.ts
```diff
@@ -0,0 +1,112 @@
...
+        };
+
+        return new Promise(resolve => {
+            this.webSocket.onopen = (): void => {
+                clearInterval(this.reconnectTimer);
+                const wrappedSocket = wrap(this.webSocket);
+                const wsConnection = createWebSocketConnection(wrappedSocket, handler.logger);
+
+                this.webSocket.onclose = (): void => {
+                    const { reconnecting } = { ...this.options };
```
tortmayr, line 82: Restructuring of options is not necessary here:
```suggestion
                    const { reconnecting } = this.options ;
```

tortmayr: Since your are extracting option props here anyways why not also add the `reconnectDelay` and `reconnectAttemtps` property?

```ts
 const { reconnecting, reconnectDelay, reconnectAttempts } = this.options ;
 ```

File: packages/protocol/src/client-server-protocol/jsonrpc/ws-connection-provider.ts [hunk 2]
```diff
@@ -0,0 +1,112 @@
...
+                            }, this.options.reconnectDelay!);
+                        }
+                    } else {
+                        handler.logger?.error('GLSPWebSocketProvider WebSocket will not reconnect - closing the connection now!');
+                    }
+                };
+
+                if (isReconnecting) {
+                    handler.logger?.warn('GLSPWebSocketProvider Reconnecting!');
+                    handler.onReconnect && handler.onReconnect(wsConnection);
```
tortmayr, line 103:
```suggestion
                    handler.onReconnect?.(wsConnection);
```

packages/protocol/src/client-server-protocol/jsonrpc/ws-connection-provider.ts
```diff
@@ -0,0 +1,112 @@
...
+                } else {
+                    handler.logger?.warn('GLSPWebSocketProvider Initializing!');
+                    handler.onConnection && handler.onConnection(wsConnection);
```
tortmayr, line 106:
```suggestion
                    handler?.onConnection(wsConnection);
```

examples/workflow-standalone/src/app.ts
```diff
@@ -57,20 +61,39 @@ async function initialize(connectionProvider: MessageConnection): Promise<void>
...
+                isReconnecting
             }
         })
     );
     actionDispatcher.dispatch(RequestTypeHintsAction.create());
     await actionDispatcher.onceModelInitialized();
     actionDispatcher.dispatch(EnableToolPaletteAction.create());
+
+    if (isReconnecting) {
+        const message = `Connection to the ${id} glsp server got closed. Connection was successfully re-established.`;
```
tortmayr, line 82: Offtopic: Dispatching `ServerStatusActions` or `ServerMessageActions` on the client side seems a bit weird.
In general, these actions are just plain status or message notifications and don't have to necessarily come from the server. Maybe we should consider a rename here to a more generic `StatusAction` and `MessageAction` or at the very least make it clear in the documentation that this actions can also be dispatched from the client side.
@planger  WDYT?

tortmayr:
Thanks Nina. Very promising change. I only had time for a code review for now. Will do some testing on Monday.

examples/workflow-standalone/src/app.ts
planger: True, I'm in favor of a rename to `StatusAction` and `MessageAction`. However, I don't consider this a high prio.

tortmayr: Me neither, nevertheless we should probably do it as part of the 2.0 Release. It' an API break after all, even if its just a minor one.

tortmayr (approved):

examples/workflow-standalone/src/app.ts
ndoschek: Sounds good, I'll create an issue for this.

ndoschek: See https://github.com/eclipse-glsp/glsp/issues/1071

PR #270, opened 2023-07-25
GLSP-1026 Resolve circular dependency between client and node-server

tortmayr:
Remove direct dependency in `@eclipse-glsp/client` and add dedicated download & start script instead. This resolves the circular dependency issues between glsp-client and server-node and unblocks us for releases.

Fixes https://github.com/eclipse-glsp/glsp/issues/1026


planger (approved):

Tested the script according to what is written in the readme.

PR #271, opened 2023-07-26
GLSP-1078: Harden property type checks if optional flag is set

ndoschek:
- Harden type checks to allow undefined value for a property if the optional flag is set
- Add respective test cases

Resolves https://github.com/eclipse-glsp/glsp/issues/1078


tortmayr (approved):
LGTM! 👍🏼

PR #272, opened 2023-07-30
GLSP-900: Introduce `GLSPModelSource`

tortmayr:
GLSP-900
- Remove the `GLSPDiagramServer` model source and introduce a new `GLSPModelSource`. In contrast to the `GLSPDiagramServer` the new model source is only responsible for sending actions to and receiving actions for the GLSP server. In addition, we get rid of the `DiagramServer` terminology which was just confusing for adopters. It no longer serves as dedicated action handler. A a cleaner separation of concerns this also enables using a generic `ModelSource` for all platform integrations i.e. we no longer have the separation between `GLSPDiagramServer´ and `GLSPTheiaDiagramServer`
- Remove the `SourceUriAware` interface (no longer needed since we only have one model source implementation)
- Introduce status-module that provides a UIExtension to handle and display `ServerStatusActions`. This overlay is a generic replacement for the status div that we typically add to the application specific diagram widgets (theia,vsode) manually. Effectively the behavior has been extracted from the `GLSPDiagramWidget` of the Theia integration
- Provide generic dirty state handling. The `EditorContextService` now also manages the dirty state and offers an event to react to dirty state changes.

GLSP-850: Refactor & improve base GLSP protocol
- Extend `onActionMessage` method of GLSPClient with optional `clientId` to be able to register handlers that are scoped to one client id/session. (no-breaking)
- Add a `initializeResult` property to the `GLSPClient` interface which caches the `initializeResult` of the first `initializeServer` invocation. Consecutive `initializeServer` calls return the cached result. (non-breaking)
- Ensure that calling `start` when the client is already running exits early and does not result in an runtime error

Part of eclipse-glsp/glsp/issues/900
Part of eclipse-glsp/glsp/issues/600


packages/protocol/src/client-server-protocol/glsp-client.ts
planger, line 1: Please also adjust the protocol doc on the [GLSP website](https://github.com/eclipse-glsp/glsp-website-source).

packages/client/src/features/status/glsp-status-overlay.ts
```diff
@@ -0,0 +1,105 @@
...
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+import { injectable } from 'inversify';
+import { AbstractUIExtension, IActionHandler, ServerStatusAction, codiconCSSClasses } from '~glsp-sprotty';
+
+@injectable()
+export class GLSPStatusOverlay extends AbstractUIExtension implements IActionHandler {
```
planger, line 20: Did we now settle to prefix things like this with GLSP?
I'm wondering whether we can just call this `StatusOverlay` now.

packages/client/src/base/model/glsp-model-source.ts
```diff
@@ -0,0 +1,149 @@
...
+ * Handles & forwards actions that are intended for the GLSP server. In addition, it handles {@link ActionMessage}s received
+ * from the server and dispatches the corresponding actions locally.
+ *
+ * Note that in sprotty a {@link ModelSource} is serving the model to the event cycle and
+ * is used to commit the local (i.e. client-side) model back to the source.
+ * However, in GLSP the update flow is reversed meaning that changes to the source model are applied
+ * on the server side and then an update is sent to the client.
+ */
+@injectable()
+export class GLSPModelSource extends ModelSource implements Disposable {
```
planger, line 63: On the long run we may want to rename that to `GLSPServerProxy`, wdyt?

planger (approved):
Great change, thank you Tobias! Just a few optional comments below that don't need any immediate action.
I tested standalone and Theia integration.

packages/client/src/base/model/glsp-model-source.ts
tortmayr: No sure. Isn't technically the `GLSPClient` our `GLSPServerProxy`? In general, I think we already have a quite confusing termonlogy with `GLSPServer, GLSPClient, ClientContribution, ClientProxy, DiagramSever etc.) so I would rather opt for a generic name here a la " This is the GLSP implementation of the Modelsource => therefore lets just call it `GLSPModelSource`)

tortmayr: Also technically would this be a `GLSPServerForwarder` an no a  proxy? 😕

tortmayr:
@planger  Thanks for the fast review. I have addressed your feedback.

packages/client/src/base/model/glsp-model-source.ts
planger: Ok, sure... yeah let's keep it like that... Also I guess it makes sense since the concept in Sprotty is called `ModelSource`.

planger (approved):

PR #273, opened 2023-08-07
GLSP-1082: Fix edge edit

tortmayr:
Update selection-mouse listener. The base mouse listener in sprotty has been reworked. This change adjusts the `RankedSelectMouseListener` to the new basae implementation and overrides it where necessary to avoid disposal of `SetEditModeActions`

Fixes https://github.com/eclipse-glsp/glsp/issues/1082


planger (approved):

PR #275, opened 2023-08-10
Add edge routing in GLSP projection view

planger:
We should compute all routes at once in order to enable edge route postprocessors and with that also the intersection detection. When only computing one route after the other we cannot detect intersections and thus don't render gaps on edge intersections.

Fixes https://github.com/eclipse-glsp/glsp/issues/1084


tortmayr (approved):
Looks good to me and edge intersection detection works again as expected!

PR #260, opened 2023-07-07
Introduce list of keyboard shortcuts

aylin-sarioglu:
This PR is part of my diploma research conducted at the [Model Engineering lab](https://me.big.tuwien.ac.at/) of TU Wien in the [Business Informatics Group](https://www.big.tuwien.ac.at/). It introduces a list of existing keyboard shortcuts.

### Search
1. Focus is on viewport.
2. Use the shortcut  `ALT+H` to display the list of keyboard shortcuts.

@ndoschek @martin-fleck-at 

Part of https://github.com/eclipse-glsp/glsp/issues/1029

PR #276, opened 2023-08-20
Introducing element navigation via keyboard

aylin-sarioglu:
This PR is part of my diploma research conducted at the [Model Engineering lab](https://me.big.tuwien.ac.at/) of TU Wien in the [Business Informatics Group](https://www.big.tuwien.ac.at/). It introduces two navigation mechanisms for iterating through the given model.
Additionally, this PR also includes a **Focus Tracker** and **Notifications**.

The **Focus Tracker** displays the current focused element with a textual indicator on the bottom left corner of the canvas. The **Notifications** are used to provide user information, e.g. activated resize or navigation modes, displayed on the canvas. Both will be triggered automatically if applicable.

### Default Navigation (following directions of relations)
1. Select an element as starting point.
2. Use the shortcut `N` to activate default navigation.
3. Use arrow keys to iterate through the model according to the directions of the given relations.

### Position-based Navigation (following x and y coordinates)

1. Select an element as starting point.
2. Use the shortcut `ALT+N` to activate position-based navigation.
3. Use arrow keys to iterate through the model according to the positions of the elements, i.e. depending on the order of the elements' x and y coordinates.

@ndoschek @martin-fleck-at

Part of https://github.com/eclipse-glsp/glsp/issues/1029


packages/client/css/navigation.css
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+.navigable-element {
+    stroke: rgb(102, 51, 51) !important;
+}
+.navigable-element .sprotty-edge.arrow {
+    fill: rgb(102, 51, 51) !important;
+    stroke: rgb(102, 51, 51) !important;
+}
```
ndoschek, line 22: I suggest to extract this color value into a variable to simplify customizations.
Also we should choose a more contrasting color, as it is rather similar to the edge select color (`.sprotty-edge.selected`).

```suggestion
:root {
    --glsp-navigation-highlight: rgb(102, 51, 51);
}

.navigable-element {
    stroke: var(--glsp-navigation-highlight) !important;
}
.navigable-element .sprotty-edge.arrow {
    fill: var(--glsp-navigation-highlight) !important;
    stroke: var(--glsp-navigation-highlight) !important;
}
```

packages/client/src/features/accessibility/element-navigation/diagram-navigation-tool.ts
```diff
@@ -0,0 +1,297 @@
...
+}
+
+enum NavigationMode {
+    POSITION = 'position',
+    DEFAULT = 'default',
+    NONE = 'none'
+}
+
+export class ElementNavigatorKeyListener extends KeyListener {
+    static readonly TIMEOUT = 5000;
```
ndoschek, line 79: Why do we need this 5s timeout here? It seems a bit long IMHO.

packages/client/src/features/accessibility/toast/messages.json
```diff
@@ -0,0 +1,16 @@
...
+    "local_navigation_mode_activated": "Position based Nav On: Navigate nearest elements using arrow keys: (↑) for above, (↓) for below, (←) for previous, (→) for next element. Press 'ALT+N' to exit.",
+    "local_navigation_mode_deactivated": "Position based Nav Off: Press 'N' for default navigation, 'ALT+N' for position based navigation."
+  },
+  "focus": {
+    "focus_on": "Currently focused: ",
+    "focus_off": "Currently no element is focused."
+  },
+  "resize": {
+    "resize_mode_activated": "Resize On: Use plus(+) and minus(-) to resize, 'CTRL'+'0' for default size. Press 'ESC' to exit.",
+    "resize_mode_deactivated": "Resize Off: Press 'ALT'+'R' for resize mode."
```
ndoschek, line 14: I guess this should be `'ALT' + 'A'`?
Also, I noticed the `Resize Off` message does never disappear on Escape not even if another message appears (e.g. by activating navigation)

packages/client/src/features/accessibility/toast/toast-tool.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    preInitialize(): void {
+        this.show(this.editorContext.modelRoot);
+    }
+}
+
+function values(obj: { [key: symbol]: ToastOptions }): ToastOptions[] {
+    return Object.getOwnPropertySymbols(obj).map(s => obj[s]);
+}
```
ndoschek, line 148: Please move this as member function into the class

packages/client/src/features/accessibility/element-navigation/local-element-navigator.ts
```diff
@@ -0,0 +1,147 @@
...
+import { BoundsAware, EdgeRouterRegistry, SConnectableElement, SEdge, SModelElement, SModelRoot, TYPES } from '~glsp-sprotty';
+import { ElementNavigator } from './element-navigator';
+import { inject, injectable, optional } from 'inversify';
+import { GLSPActionDispatcher } from '../../../base/action-dispatcher';
+import { applyCssClasses, deleteCssClasses } from '../../../base/feedback/css-feedback';
+import { SelectableBoundsAware } from '../../../utils/smodel-util';
+
+@injectable()
+export class LocalElementNavigator implements ElementNavigator {
+    navigableElementCSS = 'navigable-element';
```
ndoschek, line 25: It seems that this CSS class is not properly applied to the element currently focused in the navigation mode. Could you please doublecheck why they are not properly applied? Thanks!

ndoschek (changes requested):
Hi @aylin-sarioglu!

Thanks for this contribution! :rocket: 
I didn't have a thorough look at the code yet, but I noticed some overall issues, which I summarized below. Could you have a quick look and let me know what you think? Thank you!

- Is there a reason why the navigation modes can only be activated if an element is selected? I find this a little confusing, because I have to use the mouse to select an element and then to switch to the keyboard if I prefer that. Also, if selection is lost and navigation mode is on it cannot be restored without the mouse. Also the navigation mode cannot be exited if selection is lost. Could you have another look a t this? Also, could you think of a way to jump to the very first element for example in this case and initially?
- I would be in favor of introducing visible grouping in the shortcut overview now, as we have multiple overlapping keybindings in different contexts. Either margins between the groups, adding the group name as line or something like that.
- The Help menu cannot be opened if navigation mode is active.
- Focus tracker does not reset focus if a creation tool was selected in the toolpalette and then one of the tools in the header is activated (which aborts the creation tool):
image


Another general comment:
- Please add documentation of this accessibility feature to the [accessibility README](https://github.com/eclipse-glsp/glsp-client/tree/master/packages/client/src/features/accessibility). Could you please also add the information of the previous "help" feature (PR #260)? TIA

aylin-sarioglu:
Hi @ndoschek, Thank you for your feedback. 

> 1. Is there a reason why the navigation modes can only be activated if an element is selected? I find this a little confusing because I have to use the mouse to select an element and then switch to the keyboard if I prefer that.

Without selecting elements, activating the navigation wouldn't make sense, as its purpose is to move through those selected diagram elements.
I fixed the code a little bit, as now it is possible to focus the diagram with TAB and then the search can be called via CTRL+F so there is no need to use the mouse to select an element. The only issue here is, that it can happen that while iterating via TAB the focus can jump between the tool palette entries and the diagram. The reason for that is that the tool palette entries each have an ascending tab index, therefore the focus is jumping to different elements. The tool palette entries should have the same index.

> 2. Also, if selection is lost and navigation mode is on it cannot be restored without the mouse. 

As explained in 1.) , now you can focus on the diagram via TAB and then call the search, without the need for a mouse.

> 3. Also, the navigation mode cannot be exited if selection is lost. Could you have another look at this?  

This is fixed now.

> 4. Also, could you think of a way to jump to the very first element for example in this case and initially?

This was intentionally not added for this prototype, as there is no clear definition of what the first element of a diagram is, i.e. it could be the one positioned at the very top or left of the canvas, as this depends on the positioning of the user. But, by focusing on the diagram (which is now provided with the correction explained in 1.) and calling the search, the desired element could be selected.

packages/client/src/features/accessibility/toast/toast-tool.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+export interface ToastOptions {
+    id: symbol;
+    timeout?: number;
+    position: 'left' | 'center' | 'right';
+    message: string;
+}
+@injectable()
```
ndoschek, line 74: Could you please add a short description of this extension here? Thanks!

packages/client/src/features/accessibility/toast/toast-tool.ts
```diff
@@ -0,0 +1,155 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { injectable, inject } from 'inversify';
+import { Action, AbstractUIExtension, IActionHandler, ICommand, TYPES } from '~glsp-sprotty';
+import { IDiagramStartup } from '../../../base/model/diagram-loader';
+import { GLSPActionDispatcher } from '../../../base/action-dispatcher';
+import { EditorContextService } from '../../../base/editor-context-service';
+
+export interface HideToastAction extends Action {
```
ndoschek, line 23: Perhaps it would make sense to extract the action declarations into a separate file in this `toast` directory, wdyt?

packages/client/src/features/viewport/reposition.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    export function create(elementIDs: string[]): RepositionAction {
+        return {
+            kind: KIND,
+            elementIDs
+        };
+    }
+}
+@injectable()
```
ndoschek, line 54: Could you add a short description of this command here thanks!

ndoschek (changes requested):
Thank you very much @aylin-sarioglu for the fast update on this PR!
I added a few comments inline, but nothing major from my POV.

And I noticed one issue regarding the highlighting of navigable elements.
If you exit the navigation mode if navigable elements are highlighted, the CSS class is not reset after exiting the navigation mode (please see screenshot). Could you please doublecheck this? Thanks!

image

ndoschek (approved):
Thanks @aylin-sarioglu LGTM :+1:

PR #277, opened 2023-08-24
GLSP-1096: Update to sprotty 0.14.0

tortmayr:
Update to sprotty 0.14.0 and adapt code to conform to latest changes/API breaks. With 0.14.0 the `ToolManager API` was deprecated and it will be removed in future versions. The intention is that this API moves to GLSP as it is only used there => refactor `ToolManager API`

Fixes https://github.com/eclipse-glsp/glsp/issues/1096 
Fixes https://github.com/eclipse-glsp/glsp/issues/1094 (via sprotty update)

PR #278, opened 2023-08-25
GLSP-1088: Ensure we restore focus to the diagram

martin-fleck-at:
- UI extensions only track the active element before they are shown
- Enhance the focus tracker to provide more info on the active element
-- In Theia the tool palette stored the file explorer as active element

Requires https://github.com/eclipse-glsp/glsp-theia-integration/pull/174
Fixes https://github.com/eclipse-glsp/glsp/issues/1088


tortmayr (approved):
Changes looks good to me. Focus tracking now seems to work as expected.
FYI: This PR was based on https://github.com/eclipse-glsp/glsp-client/pull/277 which is already merged. So I rebased the PR onto the latest master.

PR #279, opened 2023-08-27
Introducing accessible tool palette, node & edge creation

aylin-sarioglu:
This PR is part of my diploma research conducted at the [Model Engineering lab](https://me.big.tuwien.ac.at/) of TU Wien in the [Business Informatics Group](https://www.big.tuwien.ac.at/). It introduces the accessible tool palette, and the Grid and Pointer functionalities, including keyboard-only alternatives for creating nodes and edges.

### Tool Palette
The shortcut `ALT + P` sets the focus on the tool palette. Afterward, the characters `a - z` select an element or `1 - 5` for the header menu options.

### Grid + Pointer
After selecting a node in the tool palette, the grid gets visible. The grid is for positioning the pointer in the screen.

The following shortcuts are usable:

1. Use keys `1 - 9` to position the pointer in the grid.
2. Use `ARROW KEYS` to move the pointer to a direction.
3. Use `ENTER` to create the node.
4. Use `CTRL + ENTER` to create multiple nodes.

### Create Nodes
1. Use `ALT + P` to focus the tool palette.
2. Use `a - z` to select a node.
3. Use `1 - 9` to position the pointer in a cell.
4. Use `ARROW KEY` to move the pointer to the correct position.
5. Create the node by using either
    - `ENTER`: Create the node und finishes the operation
    - `CTRL` + `ENTER`: Create multiple nodes

### Create Edges
1. Use `ALT + P` to focus the tool palette.
2. Use `a - z` to select an edge.
3. Type in either type or name of node for **source**.
4. Use `ENTER` to make selection.
5. Type in either type or name of node for **target**.
6. Use `ENTER` to make selection.

### Focus on Graph
1. Use  `ALT + G` to focus the graph.

@ndoschek

### DIFF
--- a/packages/client/src/features/accessibility/keyboard-pointer/keyboard-pointer.ts
+++ b/packages/client/src/features/accessibility/keyboard-pointer/keyboard-pointer.ts
@@ -1,0 +1,120 @@
+/********************************************************************************
+ * Copyright (c) 2023 Business Informatics Group (TU Wien) and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+import { inject, injectable } from 'inversify';
+import {
+    AbstractUIExtension,
+    IActionDispatcher,
+    IActionHandler,
+    SModelRoot,
+    TYPES,
+    Action,
+    TriggerNodeCreationAction
+} from '~glsp-sprotty';
+import { SetKeyboardPointerRenderPositionAction } from './actions';
+import { KeyboardPointerMetadata } from './constants';
+import { KeyboardPointerKeyboardListener } from './keyboard-pointer-listener';
+import { KeyboardPointerPosition } from './keyboard-pointer-position';
+import { KeyboardGridCellSelectedAction } from '../keyboard-grid/action';
+import { EditorContextService } from '../../../base/editor-context-service';
+import { CursorCSS } from '../../../base/feedback/css-feedback';
+
+@injectable()
+export class KeyboardPointer extends AbstractUIExtension implements IActionHandler {
+    protected _triggerAction: TriggerNodeCreationAction = {
+        elementTypeId: 'task:automated',
+        kind: 'triggerNodeCreation'
+    };
+
+    protected position: KeyboardPointerPosition = new KeyboardPointerPosition(this);
+    protected keyListener: KeyboardPointerKeyboardListener;
+    @inject(EditorContextService)
+    public editorContextService: EditorContextService;
+
+    constructor(@inject(TYPES.IActionDispatcher) protected readonly actionDispatcher: IActionDispatcher) {
+        super();
+
+        this.keyListener = new KeyboardPointerKeyboardListener(this, actionDispatcher);
+    }
+
+    get triggerAction(): TriggerNodeCreationAction {
+        return this._triggerAction;
+    }
+
+    get isVisible(): boolean {
+        return this.containerElement?.style.visibility === 'visible';
+    }
+
+    get getPosition(): KeyboardPointerPosition {
+        return this.position;
+    }
+    get getKeyListener(): KeyboardPointerKeyboardListener {
+        return this.keyListener;
+    }
+
+    id(): string {
+        return KeyboardPointerMetadata.ID;
+    }
+
+    containerClass(): string {
+        return KeyboardPointerMetadata.ID;
+    }
+
+    protected initializeContents(containerElement: HTMLElement): void {
+        containerElement.style.position = 'absolute';
+        containerElement.style.height = `${KeyboardPointerMetadata.CRICLE_HEIGHT}px`;
+        containerElement.style.width = `${KeyboardPointerMetadata.CIRCLE_WIDTH}px`;
+        containerElement.style.borderRadius = '100%';
+    }
+
+    protected override onBeforeShow(containerElement: HTMLElement, root: Readonly<SModelRoot>, ...selectedElementIds: string[]): void {
+        this.render();
+    }
+
+    handle(action: Action): Action | void {
+        if (TriggerNodeCreationAction.is(action)) {
+            this._triggerAction = action;
+        } else if (SetKeyboardPointerRenderPositionAction.is(action)) {
+            this.position.renderPosition = { x: action.x, y: action.y };
+            this.render();
+        } else if (KeyboardGridCellSelectedAction.is(action) && action.options.originId === KeyboardPointerMetadata.ID) {
+            this.position.renderPosition = action.options.centerCellPosition;
+            this.render();
+        }
+    }
+
+    render(): void {
+        if (this.containerElement !== undefined) {
+            const { x, y } = this.position.renderPosition;
+            this.containerElement.style.left = `${x}px`;
+            this.containerElement.style.top = `${y}px`;
+
+            const { status } = this.position.containableParentAtDiagramPosition(this._triggerAction.elementTypeId);
+
+            this.containerElement.style.borderStyle = 'solid';
+            this.containerElement.style.borderWidth = 'thick';
+            switch (status) {
+                case CursorCSS.NODE_CREATION: {
+                    this.containerElement.style.borderColor = 'green';
+                    break;
+                }
+                case CursorCSS.OPERATION_NOT_ALLOWED: {
+                    this.containerElement.style.borderColor = 'red';
+                    break;
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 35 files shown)


ndoschek (changes requested):
Hi @aylin-sarioglu!

Thank you for your contribution! Overall this feature is a great improvement for accessing GLSP via the keyboard! :rocket:
I had a look at the functionality but didn't look at the code thoroughly yet.
There are some issues from my POV with the usability of the accessible tool palette.
Would be great if you could have another look, thank you!

Accessible tool palette
- make the accessible tool palette mode toggleable (i.e. via pressing alt+p a second time)
- focus is always unknown in my case if I enable the accessible tool palette mode
- if a tool palette item is selected via mouse the tool palette help is enabled automatically, I would rather prefer to only enable it via the shortcut.
- shortcut help cannot be opened if accessible tool palette mode is enabled
- also any other shortcut cannot be used, e.g. element search. If I enable the delete element tool and I would like to search an element to delete, that does not work
- if create mode is exited via enter, the focus tracker tells me that the element is focused (e.g. a new automated task) but I cannot interact with it (e.g. delete it via <kbd>Del</kbd> or move it or trigger undo pf the creation). I suggest that we should immediately leave the accessible tool palette mode and select the created element.
- if the search in the tool palette is used (tool number 5), the typed keys are applied automatically to the creation tools which is not expected
- creating multiple elements: I like this feature, but I find it confusing that it is possible to create multiple elements at the exact same position without feedback. IMHO the creation target indicator (green/red circle) should update and not allow creating the element at the exact same position.
- the marquee tool (tool number 3) cannot be used via keyboard at the moment. For this we still need to use the mouse to draw a selection rectangle. Do you think this could be implemented in a reasonable amount of time or will this need more time? Alternatively, we could show some additional information to the user as toast message, wdyt?

General issues (which could also be addressed in a follow-up from my POV):
- It would be great to add a shortcut to focus the diagram (and add to help) and also initially focus diagram on startup
- The element search does not list activity nodes like fork, join, decision or merge nodes at all

aylin-sarioglu:
Hi @ndoschek , Thank you for your feedback.

Rest of your comments are fixed, some answers to the remaining remarks:

> * shortcut help cannot be opened if accessible tool palette mode is enabled,   also any other shortcut cannot be used, e.g. element search. If I enable the delete element tool and I would like to search an element to delete, that does not work

The tools that depend on a shortcut (search, help) can be only triggered if the focus is on the graph/canvas. If the focus is somewhere different (tool palette), then those key listeners are not triggered. I think the key handler is implemented like that and I'm not sure how we can overcome this problem.

> * if create mode is exited via enter, the focus tracker tells me that the element is focused (e.g. a new automated task) but I cannot interact with it (e.g. delete it via Del or move it or trigger undo pf the creation). I suggest that we should immediately leave the accessible tool palette mode and select the created element.

I tried to change the focus, however, internally somehow always the latest focused element is refocused again. I searched the code, the UIExtension restores the focus if it is hidden, however, I was not enable to change the behavior to focus on something different.


> * creating multiple elements: I like this feature, but I find it confusing that it is possible to create multiple elements at the exact same position without feedback. IMHO the creation target indicator (green/red circle) should update and not allow creating the element at the exact same position.

The target indicator has the same behavior as the mouse itself, i.e. if the mouse click allows/disallows the position then this is also applicable for the target indicator.

> * the marquee tool (tool number 3) cannot be used via keyboard at the moment. For this we still need to use the mouse to draw a selection rectangle. Do you think this could be implemented in a reasonable amount of time or will this need more time? Alternatively, we could show some additional information to the user as toast message, wdyt?

Yes, this is true, but the implementation of an alternative requires a new concept and is not possible right now. Therefore an informational toast message was added.

> * The element search does not list activity nodes like fork, join, decision or merge nodes at all

I could not recreate this. As seen in the screenshot below the activity nodes are displayed. They are not listed in the search if an edge was selected, as in that case only the elements are listed which can be chosen as source or target elements for this edge type.
nodes


For testing, the accessibility module needs to be added (as in the current version the module was removed), and to use the keyboard-tool-palette,  the default toolpalette should be removed.

ndoschek (changes requested):
Hi @aylin-sarioglu, I had another quick look and unfortunately, the behavior is a little off with the newest version.

With #280 Tobias changed the accessibility module to be default for the standalone example only. So could you please double check that the correct tool-palette is bound in the [diagram module](https://github.com/eclipse-glsp/glsp-client/blob/master/examples/workflow-standalone/src/di.config.ts) for the standalone example? Thanks!
Also, I cannot open the help view and the element search anymore, but I guess this is also linked to the above mentioned issue.
Thank you!

packages/client/src/features/accessibility/keyboard-pointer/keyboard-pointer.ts
```diff
@@ -0,0 +1,112 @@
...
+
+@injectable()
+export class KeyboardPointer extends AbstractUIExtension implements IActionHandler {
+    protected _triggerAction: TriggerNodeCreationAction = {
+        elementTypeId: 'task:automated',
+        kind: 'triggerNodeCreation'
+    };
+
+    position = new KeyboardPointerPosition(this);
+    keyListener;
```
ndoschek, line 41: They should be protected I guess?
```suggestion
    protected position: KeyboardPointerPosition = new KeyboardPointerPosition(this);
    protected keyListener: KeyboardPointerKeyboardListener;
```

packages/client/src/features/accessibility/keyboard-pointer/keyboard-pointer.ts
```diff
@@ -0,0 +1,112 @@
...
+            const { x, y } = this.position.renderPosition;
+            this.containerElement.style.left = `${x}px`;
+            this.containerElement.style.top = `${y}px`;
+
+            const { status } = this.position.containableParentAtDiagramPosition(this._triggerAction.elementTypeId);
+
+            this.containerElement.style.borderStyle = 'solid';
+            this.containerElement.style.borderWidth = 'thick';
+            switch (status) {
+                case 'NODE_CREATION': {
```
ndoschek, line 101:
```suggestion
                case CursorCSS.NODE_CREATION: {
```

packages/client/src/features/accessibility/keyboard-pointer/keyboard-pointer.ts
```diff
@@ -0,0 +1,112 @@
...
+                    this.containerElement.style.borderColor = 'green';
+                    break;
+                }
+                case 'OPERATION_NOT_ALLOWED': {
```
ndoschek, line 105:
```suggestion
                case CursorCSS.OPERATION_NOT_ALLOWED: {
```

packages/client/src/features/accessibility/keyboard-pointer/keyboard-pointer.position.ts
ndoschek, line 1: I guess the file name should be `keyboard-pointer-position.ts`?

packages/client/src/features/viewport/viewport-handler.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    protected focusGraph(): void {
+        const container = document.getElementById(this.options.baseDiv)?.querySelector(this.graphSelector) as HTMLElement | null;
+        container?.focus();
+    }
+}
+
+// https://stackoverflow.com/questions/5525071/how-to-wait-until-an-element-exists
+function waitForElement(selector: string): Promise<Element | null> {
```
ndoschek, line 49: Please move this up as member of `RestoreViewportHandler` class.

ndoschek (changes requested):
Thank you @aylin-sarioglu, this is great progress! :tada: 

Code looks good to me overall, I added some minor comments inline.

And i noticed a few issues, could you please have a look?

1. The tool palette and the selection of the grid to add elements does not work with Numpad 1-9
2. The tool palette search cannot be exited properly, on the first Esc the search string is cleared and it would be great that on second Esc or Alt+P the focus goes back to the tool palette to be able to proceed.
3. If the element search is opened, the activity nodes (decision/fork/merge/join) change their CSS color and are hard to differ from the canvas background - is this intended?

Regarding our previous discussions, I have one follow-up question, the other issues seem resolved to me :+1:

> I could not recreate this. As seen in the screenshot below the activity nodes are displayed. They are not listed in the search if an edge was selected, as in that case only the elements are listed which can be chosen as source or target elements for this edge type.

What I meant is, that they cannot be searched to be selected for example. E.g. if I want to select the first fork node in the workflow, I have no possibility to do so. They do not have a name, but I guess it would be fine for now to just list them in the results as well with their type, wdyt? They are of type `activityNode:decision|fork|join|merge `.

Thanks again!

aylin-sarioglu:
Hi @ndoschek , 

> 1. The tool palette and the selection of the grid to add elements does not work with Numpad 1-9

Fixed

> 2. The tool palette search cannot be exited properly, on the first Esc the search string is cleared and it would be great that on second Esc or Alt+P the focus goes back to the tool palette to be able to proceed.

Fixed: ALT+P is not possible, as for triggering it,  the focus must be on the graph itself. So the case via Escape was refined.

> 3. If the element search is opened, the activity nodes (decision/fork/merge/join) change their CSS color and are hard to differ from the canvas background - is this intended?

Yes, it is intended. Every element that can not be searched or is not included in the search gets transparent so that we can differentiate between the elements. In this case, adding transparency made those elements unfortunately hard to differentiate from the canvas, so I tried to fix the opacity value a little bit.

> What I meant is, that they cannot be searched to be selected for example. E.g. if I want to select the first fork node in the workflow, I have no possibility to do so. They do not have a name, but I guess it would be fine for now to just list them in the results as well with their type, wdyt? They are of type `activityNode:decision|fork|join|merge `.

Fixed

ndoschek (approved):
Thank you very much @aylin-sarioglu for the great update!
I tested again and this is a very great addition to the accessibilty modules for GLSP :rocket: 
LGTM :sparkles: :+1:

PR #280, opened 2023-08-28
Only enable a11y module in standalone example

tortmayr:
The a11y module is currently only tested for the standalone example but is configured per default in the global workflow glsp config. 
=> Remove module from global config and only enable it in the standalone example


planger (approved):
Let's activate this module individually for specific platform integrations, once they are tested there.

PR #281, opened 2023-08-29
Fix hover support for issue markers

tortmayr:
Ensure that hover support for manually constructed issue markers works as expected.
The hover mouse listener checks for `GIssueMarkers` but we configure issue markers as `SIssueMarkers` per default


planger (approved):

PR #282, opened 2023-09-04
GLSP-960: Follow-up

tortmayr:
Provide `DiagamLoaderOptions` to support dynamic configuration of the loading behavior

Follow-up for eclipse-glsp/glsp/issues/960


planger (approved):

PR #283, opened 2023-09-04
Fix download server script

tortmayr:
Ensure that old server versions are properly removed before downloading a new version.


planger (approved):

PR #284, opened 2023-09-07
GLSP-1031: Fix wrong binding in bounds-module

tortmayr:
Fixes https://github.com/eclipse-glsp/glsp/issues/1031


tortmayr:
Can also be tested with these steps:
- Resize an element
- Undo the resize operation
On master this breaks the layout. With this change it should work again

planger (approved):

PR #285, opened 2023-09-11
211: Revise TypeHints and server side feedback for creation actions

CamilleLetavernier:
- Add a new parameter to EdgeTypeHint, 'dynamic', indicating that new edges need to check with the server before allowing creation
- Add a new Request/Response to implement this check
- Update the EdgeCreationTool to check with the server when trying to create a new edge configured as Dynamic

Note:

- This PR depends on the server changes: https://github.com/eclipse-glsp/glsp-server/pull/210
  - Node Server changes are not available yet
- The mouseOver() method signature in the EdgeCreationTool makes it difficult to implement Promise-based feedback, because its signature is not a (Maybe)Promise: it's an Array of (Maybe)Promises. This means we can't actually make this method asynchronous, and using then() is also complicated. The only workaround I found to trigger a conditional feedback action was to always return a single Promise<Action>, which sometimes resolves to a No-Op Action. I don't think we have such a no-op action in GLSP at the moment, so I returned a minimal action which isn't handled anywhere. This causes a warning at runtime when this branch is executed, but the actual behavior seems correct.
- I didn't add an example usage, because I couldn't figure out a proper use case in the Workflow Example. To test this change, you need to:
  - Call `setDynamic(true)` on one of the EdgeTypeHints (e.g. `Edge` in Workflow Example)
  - Check that when trying to create a new Edge, `DefaultRequestTargetTypeHintsActionHandler` is invoked on the server, and the creation of the Edge is properly rejected. When changing the `DefaultRequestTargetTypeHintsActionHandler` implementation to return `isValid=true` instead, the Edge can be properly created
  - The intention is that client applications need to define their own logic for accepting/rejecting edge creation, when edges are defined as Dynamic; so there is no universal or generic implementation for this handler. For this reason, the Default handler always rejects creation of new dynamic edges (regular edges work as before).


packages/client/src/features/tools/edge-creation/edge-creation-tool.ts
```diff
@@ -128,33 +141,74 @@ export class EdgeCreationToolMouseListener extends DragAwareMouseListener {
...
                     this.allowedTarget = this.isAllowedSource(newCurrentTarget);
                 } else if (!this.isTargetSelected()) {
-                    this.allowedTarget = this.isAllowedTarget(newCurrentTarget);
-                }
-                if (this.allowedTarget) {
-                    const action = !this.isSourceSelected()
-                        ? cursorFeedbackAction(CursorCSS.EDGE_CREATION_SOURCE)
-                        : cursorFeedbackAction(CursorCSS.EDGE_CREATION_TARGET);
-                    return [action];
+                    // Temporarily mark the target as invalid while we check for a proper result,
```
tortmayr, line 152: Could be extracted into a submethod for better extensibility.

packages/client/src/features/tools/edge-creation/edge-creation-tool.ts
```diff
@@ -128,33 +141,74 @@ export class EdgeCreationToolMouseListener extends DragAwareMouseListener {
...
+                    // so a fast-clicking user doesn't get a chance to create the edge in the meantime.
+                    this.allowedTarget = false;
+                    const actions = this.isAllowedTarget(newCurrentTarget).then(allowedCheck => {
+                        // Make sure we didn't change the target element while
+                        // checking for valid target.
+                        if (allowedCheck.targetElement === this.currentTarget) {
+                            this.allowedTarget = allowedCheck.allowed;
+                            return this.updateEdgeFeedback();
+                        }
+                        // FIXME Is there a proper no-op action? We can't return <undefined> or [] here
```
tortmayr, line 162: I wondering whether we really need to add this to the return result of the `mouseOver` method. We have access to the action dispatcher, so couldn't we simply dispatch the action returned from `this.updateEdgeFeedback()` directly with the dispatcher?

packages/protocol/src/action-protocol/element-type-hints.ts
```diff
@@ -133,3 +143,70 @@ export namespace SetTypeHintsAction {
...
+    /**
+     * The ID of the edge target element to check.
+     */
+    targetElementId: string;
+}
+
+export namespace RequestCheckEdgeTargetAction {
+    export const KIND = 'requestCheckEdgeTarget';
+
+    export function create(
```
tortmayr, line 192: For actions with more than two parameters we typically encapsulate the parameters in an `options` object:
```ts
```suggestion
     export function create(options: {
        sourceElement: SModelElement | string;
        targetElement: SModelElement | string;
        edgeTypeId: string;
    }): RequestCheckEdgeTargetAction {
        return {
            kind: KIND,
            edgeTypeId: options.edgeTypeId,
            sourceElementId: getElementId(options.sourceElement),
            targetElementId: getElementId(options.targetElement),
            requestId: ''
        };
    }
}
```

packages/protocol/src/action-protocol/element-type-hints.ts
```diff
@@ -73,9 +74,18 @@ export interface EdgeTypeHint extends TypeHint {
...
     readonly targetElementTypeIds: string[];
+
+    /**
+     * Indicates whether this type hint is dynamic or not. Dynamic edge type hints
+     * require an additional runtime check before creating an edge, when checking
+     * source and target element types is not sufficient.
+     *
+     * @see {@link RequestCheckEdgeTargetAction}
+     */
+    readonly dynamic?: boolean;
```
tortmayr, line 88: Just to make sure I got this right: So we basically now support three main usecases,right?:
-  Allowed target detection via a set of allowed source/target ids that are queried one time during startup
-  Mixed. i.e. first detect via allowed source/target id array and then trigger the dynamic server check
- Pure dynamic. I.e. the source/target id arrays are empty and we forward every target check to the server

packages/client/src/features/tools/edge-creation/edge-creation-tool.ts
```diff
@@ -128,33 +141,74 @@ export class EdgeCreationToolMouseListener extends DragAwareMouseListener {
...
+    }
+
     protected isAllowedSource(element: SModelElement | undefined): boolean {
         return element !== undefined && isConnectable(element) && element.canConnect(this.proxyEdge, 'source');
     }
 
-    protected isAllowedTarget(element: SModelElement | undefined): boolean {
-        return element !== undefined && isConnectable(element) && element.canConnect(this.proxyEdge, 'target');
+    protected async isAllowedTarget(element: SModelElement | undefined): Promise<AllowedTargetCheck> {
+        let allowed = element !== undefined && isConnectable(element) && element.canConnect(this.proxyEdge, 'target');
```
tortmayr, line 192: I think this code does not work for purley dyanmic usecases i.e. we don't pass any allowed source/target ids
AFAIK currently the `canConnect` method of elements is overwritten by our typeHintsModule (https://github.com/eclipse-glsp/glsp-client/blob/256d64ddbb4f5596d75df654a8a93502e314d092/packages/client/src/features/hints/type-hints.ts#L124)
Am I missing something here or would `allowed` always be false if dynamic target resolution is activated? (Because can connect should be false in that case)

tortmayr (changes requested):
Thanks @CamilleLetavernier.
I only had time to do a preliminary code review for now, but I will also test the changes later today

packages/protocol/src/action-protocol/element-type-hints.ts
CamilleLetavernier: No, there are only 2 cases supported: fixed and mixed.

To define a Pure dynamic case, you'd need to define an edge that accepts all possible source and target types, and make it dynamic. The static check (sourceType/targetType) is always done before the dynamic check, so if that fails, the dynamic check is skipped entirely.

packages/client/src/features/tools/edge-creation/edge-creation-tool.ts
CamilleLetavernier: That would be an option indeed; I tried to stick to the defined API to return the feedback actions. But since the mouse tool only delegates to the ActionDispatcher, I guess it'd be fine to do it directly in the listener.

packages/client/src/features/tools/edge-creation/edge-creation-tool.ts
CamilleLetavernier: Indeed; the "purely dynamic" use case is not supported. The dynamic check always comes after the regular type hint check, so if the regular check fails, the edge cannot be created.

tortmayr:
@martin-fleck-at  I have taken over this draft from Camille and have finalized the PR could you please have a look at it?

packages/client/src/features/tools/edge-creation/edge-creation-tool.ts
```diff
@@ -131,30 +143,61 @@ export class EdgeCreationToolMouseListener extends DragAwareMouseListener {
...
+        }
+        if (!this.isDynamic(this.proxyEdge.type)) {
+            return true;
+        }
+        const sourceElement = this.source ?? element;
+        const targetElement = this.source ? element : undefined;
+
+        this.pendingDynamicCheck = true;
+        // Request server edge check
+        this.actionDispatcher
```
martin-fleck-at, line 184: I wonder whether we should provide some feedback on the cursor while we are waiting for the result (e.g., wait cursor) as we do not know exactly how long the calculation may take.

packages/protocol/src/action-protocol/element-type-hints.ts
```diff
@@ -133,3 +147,119 @@ export namespace SetTypeHintsAction {
         };
     }
 }
+
+/**
+ * Send a Request to the server to check if an element is a valid target
```
martin-fleck-at, line 152: This doc is confusing imho. This action is not to check whether an element is a valid target, it is to check whether the edge information that we provide already prohibits the creation of a valid model. I'm mainly concerned about the use of `target` in the context of the edge here because I think it adds a lot of confusion. In the tool it is okay because there we also have the target of a click/hover/etc. but here I think we should find something more exact. What do you think?

martin-fleck-at (changes requested):
I did a quick test on this and the changes look good to me and work as expected. I do have some minor comments that could be integrated but nothing related to the core functionality. Thank you!

packages/protocol/src/action-protocol/element-type-hints.ts
```diff
@@ -133,3 +147,119 @@ export namespace SetTypeHintsAction {
...
+            kind: KIND,
+            edgeType: options.edgeType,
+            sourceElementId: getElementId(options.sourceElement),
+            targetElementId: options.targetElement ? getElementId(options.targetElement) : undefined,
+            requestId: options.requestId ?? ''
+        };
+    }
+}
+
+function getElementId(element: SModelElement | string): string {
```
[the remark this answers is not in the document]
tortmayr: The `getElementTypeId` of `smodel-util.ts` is more or less the exact same function. However, it lives in the client so we cannot import it here (protocol).  I don't think it makes sense to introduce a model-util.ts in the protocol package for just one function so I opted for code duplication here.

martin-fleck-at (approved):
LGTM!

tortmayr (approved):

PR #287, opened 2023-09-12
GLSP-1116 Revise model loading

tortmayr:
# GLSP-1116 Revise model loading
- Refactor diagram loader to properly dispatch the `RequestModelAction` as request action
   - Remove dispatching of temporary empty set model action and instead call `actionDispatcher.initialize()`  earlier which also dispatches an empty set model action under the hood  
- Rework `ModelInitializationConstraint`
   - Remove coupling with `ActionDispatcher` the initialization state is now controlled by the diagram loader
   - Provide `onInitialized` event and a callback function for on-time invocation `onceInitialized`
   - Provide test cases
- Add `dispatchOnceModelInitialized` utility function to action dispatcher
- Ensure that type hints are  requested after the model has been initialized

Part-of: https://github.com/eclipse-glsp/glsp/issues/1116
Part-of: https://github.com/eclipse-glsp/glsp/issues/606

# GLSP-1117: Remove need for explicit definition of client actions
-Extend `initializeClientSession` request to also specific the set of client action kinds. This way the server knows which actions should be sent to the client
-Adapt `GLSPModelSource´ to retrieve the client actions before sending the `initializeClientSession` request
- Add customized `GLSPActionHandlerRegistry` which provides a query method to retrieve all handled action kinds

Part of eclipse-glsp/glsp/issues/1117

# GLSP-1071: Rename ServerStatus/ServerMessage action
Part of https://github.com/eclipse-glsp/glsp/issues/1071

 Requires: https://github.com/eclipse-glsp/glsp-server/pull/211


packages/client/src/base/model-initialization-constraint.ts
planger, line 1: I believe some projects have customized the model initialization constraint to wait for additional custom actions before a model can be considered initialized. How would these scenarios be handled now?

packages/client/src/base/model-initialization-constraint.ts
```diff
@@ -23,70 +23,72 @@ import { Action, Deferred, InitializeCanvasBoundsAction, SetModelAction, UpdateM
...
+ * from the server.
  */
 @injectable()
-export abstract class ModelInitializationConstraint {
-    protected completion: Deferred<void> = new Deferred();
-    protected completed = false;
-
-    get isCompleted(): boolean {
-        return this.completed;
+export class ModelInitializationConstraint implements Disposable {
```
planger, line 31: I wonder if `ModelInitializationConstraint` is now still a good name. Shouldn't it now be `ModelIinitializationStatus` or something like that?

planger (commented):
This looks great! I just have a general and a naming question inline.

packages/client/src/base/model-initialization-constraint.ts
tortmayr: That's a good point. I wasn't aware of that usecase.
In that case we should probably keep the old approach (i.e. control the constraint via action dispatcher) and just to  slight refactoring to support events.
And in the diagram loader we should probably use the initialzation constraint and only trigger the `preModelInitialized` hooks once the model constraint is completed.
This should behave exactly the same as the current request model action -> re dispatch approach but is more flexible.

planger (approved):

tortmayr:
@planger I have also added change for GLSP-1117 and GLSP-1071 to this PR. Since these are also changes that require updatse across all integration projects I would rather do this all in one go instead of having to create multiple PRs per repository.

Could you please re-review?

planger (approved):

PR #288, opened 2023-09-20
GLSP-1116: Folllow-up fixes

tortmayr:
- Refactor diagramLoader+modelsource: Avoid concurrency/timing issues between diagramloader & modelsource. In some cases the loader did send actions before the model source was fully configured. Now the loader explictly configures the model source before dispatching the model request actions.
- Ensure that client session is disposed if `GlspModelSource` is disposed
- Ensure that the `dispatch` method correctly awaits the super result
- Fix typing of `requstUntil`. This method can actually also return undefined if the timeout was reached  and `rejectOnTimeout` is false + adapt affected code
- Refactor `handleAction` method to allow dispatching of received responses 
- Fix typing for Request actions (see https://github.com/eclipse-sprotty/sprotty/issues/378)
- Add test cases

- Remove deprecated code
- Fix return type of `ExportSvgAction.is`

Follow up for eclipse-glsp/glsp/issues/1116

Test with:
Java-server: master 
Node-server: https://github.com/eclipse-glsp/glsp-server-node/pull/58


packages/client/src/base/action-dispatcher.spec.ts
```diff
@@ -0,0 +1,87 @@
+/********************************************************************************
```
martin-fleck-at, line 1: Yay, tests!

packages/client/src/base/action-dispatcher.ts
```diff
@@ -69,11 +69,11 @@ export class GLSPActionDispatcher extends ActionDispatcher {
                 this.timeouts.delete(action.responseId);
             }
 
-            // we might have reached a timeout, so we simply drop the response
+            // Check if we have a pending request for the response.
+            // If not the  we clear the responseId => action will be dispatched normally
```
martin-fleck-at, line 73: That is quite the change in behavior and I believe it needs to be documented somewhere :thinking:

martin-fleck-at (approved):
Tested and looks fantastic to me. Those small changes definitely add up to quite an improved version, thank you!

packages/client/src/base/action-dispatcher.ts
tortmayr: Will be added to the changelog (I will do a collective update before the 2.0.0 release).
Anyhow, it's more of a bug fix. With the previous implementation it was not possible to dispatch a `Response` action with a set resposne id when using the `GLSPActionDispatcher`. But this is possible (and also documented in the jsdoc of request & requestUntil) with the super implementation from sprotty.

(Note that the response id clearing is optional, and is more or less just there to avoid an unnecesary warning log in the super implentation)

PR #291, opened 2023-10-09
GLSP-865: Align client model namespace with server model

tortmayr:
Also export all SModel elements (and releated) as aliased/augmented classes in the `GModel` namespace. This means we now can consistently use the GModel syntax on both server and client side.

Move package-internal glsp-sprotty reexport into dedicated package. The package-internal reexport required overhead with `tsc-alias` which did not play well with the test setup and ide support (auto-import). Using a dedicated package also improved DevXP in glsp-client as auto-import now longer suggests imports from sprotty or sprotty-protocol.

Sprotty and GLSP use different definitions of `Action`. While they are fully compatible this kind of breaks the  IDE support for API that uses `Action`. e.g. Implementing an `IActionHandler` would have always auto-imported the wrong `Action` type (from sprotty-protocol). To bypass this issue we extend the sprotty core API classes in question and reexport them instead (see action-override.ts).

Also: 
- reexport sprotty modules that are used in glsp as `FeatureModules`.
- Refactor `Argumentable` interface to a model feature and rename to `ArgsAware` and ensure that the argsFeature gets added to each element per default
- Augment all actions directly provided by sprotty-protocol to provide a typeguard via namespace
- Move all reusable autocomplete compontens into base package
- Remove "baseUrl" from tsconfig to avoid breaking "src" auto imports in vscode

Part of eclipse-glsp/glsp/issues/865


martin-fleck-at (changes requested):
Change looks good to me overall but there are a few classes still exported with `SModel...`. Other than that, I could not identify any issues, thank you Tobias!

packages/protocol/src/index.ts
```diff
@@ -13,31 +13,6 @@
...
-    Action.hasKind(object, SetViewportAction.KIND) &&
-    hasStringProp(object, 'elementId') &&
-    hasObjectProp(object, 'newViewport') &&
-    hasBooleanProp(object, 'animate');
-
-SetBoundsAction.is = (object: any): object is SetBoundsAction =>
-    Action.hasKind(object, SetBoundsAction.KIND) && hasObjectProp(object, 'bounds');
 
 // Partial reexport of sprotty-protocol
 export { Viewport } from 'sprotty-protocol/lib/model';
```
martin-fleck-at, line 18: I believe this file also exports `SModelIndex`, one of the few classes still being exported with the `SModel`-prefix. Similar to `SModelFactory` and `SModelRegistry`.

tortmayr: Good catch. We don`t need `SModelIndex` from protocol at all. This is an index of the serialize format (schema) which we don't use.
Regarding the other `SModel`-prefix exports. These are on purpose
- SEdgeImpl, SGraphImpl, SModelRegistry and SModelFactory are exported without alias because we have dedicated `G` subclasses for them in the client.

I have added inline comments to make this clearer.

martin-fleck-at: I see, I was just surprised that the code completion in the client package offered me those types instead of the expected G types but if they are only defined in the client, that makes sense. Thank you for clarifying!

packages/glsp-sprotty/src/index.ts
```diff
@@ -73,7 +73,9 @@ export {
     SModelElementConstructor as GModelElementConstructor,
     SModelElementRegistration as GModelElementRegistration,
     IModelFactory,
+    // exported without alias we extend it in glsp-client to `SModelFactory`
```
martin-fleck-at, line 76: Could you show me where that is happening? For some reason I cannot find it. It seems we have our own definition of `GModelFactory` and `SModelFactory` is not really used or am I just too blind to find it?

tortmayr: Sorry my bad. You are right we only extend the `SModelRegistry` on the client. Confusion from my side, we can simply alias `SModelFactory`

martin-fleck-at (approved):
Great, now the changes look good ot me!

PR #293, opened 2023-10-13
Bugfixes for 2.0.0

tortmayr:
- Add copy-paste support for standalone via keyboard short cuts
- Trigger auto layout operation via keyboard short cut (Shift+ctrl+L)
- Ensure that the viewport handler does not steal focus when integrated in an application frame (i.e. Theia)

- Make context menu listener selection aware. If the context menu was triggered for a selectable target (or parent):
 do nothing if the element is already selected, otherwise set the selection to the selectable element
Remove the current selection if the context menu element is not selectable. This behavior is consistent with other graphical tools like DrawIO, Google drawings, Papyrus etc.

### DIFF
--- a/packages/client/src/features/copy-paste/copy-paste-standalone.ts
+++ b/packages/client/src/features/copy-paste/copy-paste-standalone.ts
@@ -1,0 +1,96 @@
+/********************************************************************************
+ * Copyright (c) 2023 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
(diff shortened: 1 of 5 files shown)


packages/client/src/features/copy-paste/copy-paste-standalone.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @preDestroy()
+    dispose(): void {
+        this.toDispose.dispose();
+    }
+}
+
+@injectable()
+export class CopyPasteKeyListener extends KeyListener {
```
martin-fleck-at, line 83: This is not used anywhere which is probably why the cut/copy/paste is not working in the standalone case.

martin-fleck-at (changes requested):
Thank you! Everything seems to work except the cut/copy/paste in the standalone example. I assume we are missing the listener binding somewhere, could you please double check?

packages/client/src/features/copy-paste/copy-paste-standalone.ts
tortmayr: No, this class can be removed and was committed on accident.
It looks like adding the copy event listeners to the baseDiv was a mistake. They are not triggered when child elements are selected.
However, copy/paste works if you do a select all.... I have to look into this

martin-fleck-at (approved):
Thank you for the quick update. I re-tested and everything is working as expected now, great work!

PR #295, opened 2023-11-03
Keyboard move and resize improvements

planger:
* Show client-side feedback without animation
* Trigger server operation with a debounce delay
* Respect snap modifier (ALT) for move
* Respect movement restrictor for move

Can only be tested with the standalone workflow example!

Fixes https://github.com/eclipse-glsp/glsp/issues/1156

### DIFF
--- a/packages/client/src/features/accessibility/move-zoom/move-handler.ts
+++ b/packages/client/src/features/accessibility/move-zoom/move-handler.ts
@@ -14,24 +14,34 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import { inject, injectable } from 'inversify';
-import { throttle } from 'lodash';
 import {
     Action,
     ChangeBoundsOperation,
+    DisposableCollection,
+    ElementAndBounds,
+    ElementMove,
+    GModelRoot,
     IActionDispatcher,
     IActionHandler,
     ICommand,
+    ISnapper,
+    MoveAction,
     Point,
-    GModelRoot,
     SetViewportAction,
     TYPES,
     Viewport,
     findParentByFeature,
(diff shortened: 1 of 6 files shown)


martin-fleck-at (commented):
The experience is so much better now, great work Philip! It is so smooth and I especially like the moving of the view port!

I only have one question regarding the grid: I'm not sure how it is supposed to work but in my mind it was just an imaginative grid across the whole diagram. So when I move a node a few pixels using ALT but then move it with the regular key arrow I would have expected it to be aligned with the grid and not just jump one grid width/height. The mouse tool always snaps to the next grid-aligned position:

move-tool

So maybe that is not the intention of the keytool at all but I just wanted your input on that in case we need to use the snapper not just for the grid but also to calculate a valid position.

planger:
@martin-fleck-at Thank you very much! Nice find with the grid. This was indeed wrong in the original implementation. I fixed this with a72f6ff.

martin-fleck-at (approved):
The behavior is even better now, thank you very much for the fast update!

I noticed one small oddity which might be fine anyway but I'd like to get your input: We properly snap to the grid when not using the ALT key now but it we do it on both axis, so when I do ALT to move a bit to the right and then use the Down Arrow without ALT, it not only moves the element down but also to the left or right depending on the proper snap/grid location. Maybe this is expected anyway, I was just a bit surprised that suddenly the element moved on another axis as well.
snap-x-y

planger:
Right, that's because we pass it through the snapper, which isn't aware of the axis on whiich the user originally moved. We could reset the axis that the user didn't impact. But I guess this is fine for now. WDYT?

martin-fleck-at:
@planger Agreed, that's why I also approved the PR. I'm not sure how other tools are handling this but if it ever comes to it, we can simply change the behavior. Both seem "semantically" correct to me ;-)

PR #296, opened 2023-11-03
Allow setting `ViewerOptions` and set default zoom limits

planger:
Sprotty introduced zoom limits, so we should allow setting them.
This change allows to optionally specify all `ViewerOptions` in `configureDiagramOptions` and sets very generous `zoomLimits` by default. This should also mitigate the bug when you zoom beyond `Number.maxValue`.

### DIFF
--- a/packages/client/src/base/model/diagram-loader.ts
+++ b/packages/client/src/base/model/diagram-loader.ts
@@ -14,7 +14,6 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import { inject, injectable, multiInject, optional, postConstruct } from 'inversify';
 import {
     AnyObject,
     ApplicationIdProvider,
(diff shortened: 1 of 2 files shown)


tortmayr:
I'm not a fan of exposing partial IViewer options via the `IDiagramOptions` interface. With this approach
there is a potential mismatch if an adopter overrides the `zoomLimit` value in the viewer options. (IDIagramOptions than still holds the original value)

In my opinion the  `configureDiagramOptions` method should take options of type `IDiagramOptions & Partial<ViewerOptions>`.
Alternatively we could also make the `DiagramOptions` interface a subinterface of `ViewerOptions` and rebind the `ViewerOptions` service identfier.

planger:
Good comment, I agree that this approach wasn't very versatile. However combining `IDiagramOptions` and `Partial<ViewerOptions>` isn't very nice either because we'd have to explicitly pass the set `ViewerOptions` to avoid inadvertantly adding `IDiagramOptions` properties into the options.

Thus, I opted to introduce an optional second argument. WDYT?

tortmayr (approved):
Thanks. Looks good to me. I like the approach with a second options argument for the viewer options. 👍🏼

PR #298, opened 2023-11-06
Restore Projection bars

martin-fleck-at:
Relates https://github.com/eclipse-glsp/glsp/issues/1122


tortmayr (commented):
Looks good to me. Projection bars now work again as expected.
However, in the standalone example the styling is not ideal.
The scrollbar borders are hardly visible and maybe we could move the focus tracker UI Widget a little bit up so that it does not interfere with the projection bars.

tortmayr (approved):
LGTM! 👍🏼

PR #302, opened 2023-11-13
Fix download example-server script

tortmayr:
Ensure that the server download script downloads the latest version of the bundled workflow-server.
(The 2.0.0 version was bundled incorrectly and this was hotfixed with 2.0.1)


martin-fleck-at (approved):
LGTM!

PR #307, opened 2023-11-29
GLSP-1181: Fix support for simultaneously open diagrams

tortmayr:
Fixes  https://github.com/eclipse-glsp/glsp/issues/1181

Also: Update changelog to reflect latest changes

### DIFF
--- a/packages/protocol/src/client-server-protocol/jsonrpc/base-jsonrpc-glsp-client.ts
+++ b/packages/protocol/src/client-server-protocol/jsonrpc/base-jsonrpc-glsp-client.ts
@@ -36,6 +36,11 @@ export class BaseJsonrpcGLSPClient implements GLSPClient {
         return this.onServerInitializedEmitter.event;
     }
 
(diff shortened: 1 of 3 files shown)


packages/protocol/src/client-server-protocol/jsonrpc/base-jsonrpc-glsp-client.ts
```diff
@@ -91,6 +96,9 @@ export class BaseJsonrpcGLSPClient implements GLSPClient {
         try {
             this.state = ClientState.Starting;
             const connection = await this.resolveConnection();
+            connection.onNotification(JsonrpcGLSPClient.ActionMessageNotification, msg =>
```
martin-fleck-at, line 99: nit: Should we add that to the `doCreateConnection` method where we already add handlers for `onError` and `onClose`? I think it would be a better fit.

martin-fleck-at (changes requested):
Looks good to me and works well, thank you for the fix!

martin-fleck-at (approved):
LGTM!

PR #309, opened 2023-12-16
Add generic args property to all operations

tortmayr:


martin-fleck-at (commented):
@tortmayr Change looks fine to me but could you elaborate on why we need a generic map for all operations? They can only contain JSON primitives (string, number boolean) but no objects there benefit is a bit restricted.

tortmayr:
The question is rather why only have it for certain operation imo it make sense to have the possibility to add generic args to all operations. A concrete usecase that I have in one project is that i need to pass the information whether a route is routed by the user (i.e. a routing point has been dragged manually) or default routed.
The restriction to primitive types is unfortunate and we should look into that. (I think we have to make sure that non-primitive types can be converted correctly in the Java case) but  that's something that we look into in a follow-up PR

martin-fleck-at (commented):
@tortmayr Wouldn't it be enough to have a custom `ChangeRoutingPoints` operation that still uses the same operation kind but has extended arguments? I just want to avoid that we use a generic map as a default to add a lot of semantic to an operation instead of actually extending it and therefore gaining more type safety. I'm not strictly against adding the map and I'm sure there are use cases where its usage is justified, it just shouldn't be the default extension mechanism. Any answer is fine and I'll approve the PR ;-)

tortmayr:
I think the philosophy is comparable to what we offer with the generic args Map in `GModelElement`.
If you just want to add one, or two simple properties your can use the map (and don't have to declare custom extended actions on both sides)  for more complex additions extending is recommended.
In any way I would like to clean up the current inconsistent state where we have a generic args map for a couple of operations (create operations, Reconnect, ChangeContainer...) but not all of them.

In the end I think we should leave it up to adopter which strategy they want to use. 
I think the limitation to primitive values (even though its a technical one to support parsing in the Java case) comes in handy here. This way we indirectly automatically guide the adopter to use extension for more complex use cases.

The question whether we should support generic primitive args in the first place is something that definitely can be discussed, but since it's now already part of the core API we should opt for supporting it consistently.

martin-fleck-at (approved):
Great, I agree. Thank you for the discussion. LGTM!

PR #310, opened 2024-01-09
Allow optional actions to avoid console errors

martin-fleck-at:
See https://github.com/eclipse-glsp/glsp/discussions/1203

You can test it by removing the `helperLineModule` from the `workflow-diagram-module.ts` to see how it behaves before and after.


tortmayr (approved):
Thanks! Code looks good to me and seems to work as expected.

PR #312, opened 2024-01-11
Make HelperLine module configurable from central options

martin-fleck-at:


tortmayr (approved):
Changes look good to me. Thanks Martin!

PR #313, opened 2024-01-18
GLSP-1194: Do not send out selection changes if only the root changed

martin-fleck-at:
When we receive a model update and the selection state (ids of selected elements) stays the same, we should not send out a selection changed event. Code interested in model updates can listen to root changes manually. This also prevents Theia from permanently updating the selection if only the root element was changed.

Fixes https://github.com/eclipse-glsp/glsp/issues/1194

This can be tested with a Node server that sends an update every second, see https://github.com/eclipse-glsp/glsp-server-node/tree/issues/1194_test


tortmayr (approved):
LGTM and works as expected!

PR #317, opened 2024-02-02
GLSP-1223 Align sprotty-protocol version

tortmayr:
#### What it does

Ensure that 1.0.0 of sprotty-protocol is used in all client pacakges Part of https://github.com/eclipse-glsp/glsp/issues/1223

#### How to test

Nothing particular to test. Building + starting the example application should still work as before

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #321, opened 2024-03-04
Avoid dependency on 'chai' through exported index

martin-fleck-at:
#### What it does

Fixes https://github.com/eclipse-glsp/glsp/issues/1269

#### How to test

Verify through another package that does not have chai but still imports types and functions from the main package that it properly works.

#### Changelog


tortmayr (approved):
LGTM!

PR #322, opened 2024-03-04
Ensure we apply feedback actions already on SetModel

martin-fleck-at:
#### What it does

Ensure that feedback actions are already applied on set-model and not only on update-model.
Fixes https://github.com/eclipse-glsp/glsp/issues/1239

#### How to test

In the workflow example, you can add a startup listener that adds feedback early on, e.g.,
```
import { IDiagramStartup, IFeedbackActionDispatcher, MaybePromise, SelectAction, TYPES } from '@eclipse-glsp/client';
import { inject, injectable } from 'inversify';

@injectable()
export class StartupFeedback implements IDiagramStartup {
    @inject(TYPES.IFeedbackActionDispatcher) protected feedbackDispatcher: IFeedbackActionDispatcher;

    preRequestModel(): MaybePromise<void> {
        this.feedbackDispatcher.registerFeedback(this, [SelectAction.setSelection(['task0'])]);
    }
}
```

With this (and the correct binding in the module) you should see that we have a select feedback registered and applied even if all you do is open the model.

#### Changelog
- [x] This PR introduces a breaking change (if yes, provide more details below for the changelog and the migration guide)

Method `getFeedbackCommands` was moved from `FeedbackAwareUpdateModelCommand` to `IFeedbackEmitter` for re-use. `FeedbackAwareUpdateModelCommand.getPriority` was replaced by a generic `Prioritized.getPriority`.


packages/client/src/base/feedback/set-model-command.ts
```diff
@@ -0,0 +1,79 @@
...
+        actionHandlerRegistryProvider().then(registry => (this.actionHandlerRegistry = registry));
+    }
+
+    override execute(context: CommandExecutionContext): GModelRoot {
+        const root = super.execute(context);
+        this.applyFeedback(root, context);
+        return root;
+    }
+
+    protected applyFeedback(newRoot: GModelRoot, context: CommandExecutionContext): void {
```
tortmayr, line 50: To avoid code duplication we should probably move extract the `applyFeedback` and `getFeedbackCommands` methods into functions that can then be reused in both the `UpdateModel` and `SetModel` command handlers.

tortmayr, line 50: To avoid code duplication we should probably move extract the `getFeedbackCommands` methods into a shared function that can then be reused in both the `UpdateModel` and `SetModel` command handlers.

martin-fleck-at: Good point, I move the functionality directly into the feedback action dispatcher.

packages/client/src/base/priority.ts
```diff
@@ -0,0 +1,50 @@
...
+import { AnyObject, hasNumberProp } from '@eclipse-glsp/sprotty';
+
+/**
+ * A common interface for objects that can be orderable by a priority.
+ */
+export interface Prioritized {
+    priority: number;
+}
+
+export namespace Prioritized {
```
tortmayr, line 26: We already have a `Ranked` API that more or less servers the same purpose.
Maybe it would be possible to combine those two APIS or at least deprecate on of them so that we don`t have 
two APIS for the same purpose in parallel?

packages/client/src/base/feedback/feedback-action-dispatcher.ts
```diff
@@ -57,6 +71,16 @@ export interface IFeedbackActionDispatcher {
...
+
+    /**
+     * Retrieves all commands based on the registered feedback actions, ordered by their priority (highest priority first).
+     */
+    getFeedbackCommands(): Command[];
+
+    /**
+     * Applies all current feedback commands to the given command execution context.
+     */
+    applyFeedbackCommands(context: CommandExecutionContext): Promise<void>;
```
tortmayr, line 83: This introduces a (minor) API break for potential `IFeedbackActionDispatcher` implementations.
At the very least we should mention it in the changelog/migration guide.

packages/client/src/base/feedback/update-model-command.ts
```diff
@@ -59,38 +55,13 @@ export class FeedbackAwareUpdateModelCommand extends UpdateModelCommand {
...
-        }
-
+        // Create a temporary context which defines the `newRoot` as `root`
+        // This way we do not corrupt the redo/undo behavior of the super class
+        const tempContext: CommandExecutionContext = { ...context, root: newRoot };
+        this.feedbackActionDispatcher?.applyFeedbackCommands(tempContext);
         return super.performUpdate(oldRoot, newRoot, context);
     }
 
-    protected getFeedbackCommands(registry: ActionHandlerRegistry): Command[] {
```
tortmayr, line 77: This is a soft API break for adopters that have a customization of the `FeedbackAwareUpdateModelCommand`.
Should be mentioned in the changelog.

martin-fleck-at: I do have that written in the PR description, should I also already note it down somewhere in the changelog?

packages/client/src/base/feedback/feedback-action-dispatcher.ts
martin-fleck-at: I could mark the method as optional, that would allow for backwards compatibility, would you prefer that?

packages/client/src/base/priority.ts
martin-fleck-at: Yeah, I saw that and I wasn't entirely sure how to best go forward with this. They also seem to have reverted semantics, i.e., a lower rank means you are used before elements with a higher rank but for priority it is kind of the reverse, a high priority means you are before other elements. Sure, semantically they are the same but it was hard to choose which to deprecate if at all.

martin-fleck-at: I'll have another look at this tomorrow.

packages/client/src/base/feedback/feedback-action-dispatcher.ts
tortmayr: No, I don't think making it optional is the right way to go here.
I think it this case it is fine that we introduce a soft API break. We should just mention it in the changelog.
Also it's probably very unlikely that somebody uses a complete custom `IFeedbackActionDisptacher` implementation.
If adopters have custom code they probably extended the `FeedbackActionDispatcher` base class anyways. 
So in practice this break should have little to no impact.

packages/client/src/base/priority.ts
martin-fleck-at: I now use `rank` everywhere and introduced a backwards compatible property.

martin-fleck-at:
@tortmayr Locally I seem some exceptions I still need to have a look at. Will report back.

martin-fleck-at:
@tortmayr I managed to resolve the dependency issue by using the contribution provider approach from Theia.

packages/protocol/src/utils/contribution-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+                // eslint-disable-next-line no-null/no-null
+                currentContainer = recursive === true ? currentContainer.parent : null;
+            }
+            this.services = currentServices;
+        }
+        return this.services;
+    }
+}
+
+export type Bindable = interfaces.Bind | interfaces.Container;
```
tortmayr, line 61: We should remove this and reuse the `BindingContext` syntax that we use in other inversify utility functions as well.

packages/protocol/src/utils/contribution-provider.ts
```diff
@@ -0,0 +1,79 @@
+/*******************************************************************************
```
tortmayr, line 1: For consistency`s sake we should use one of the patterns already in use for code based  on third-party sources
Either (websocket-connection.ts):
```ts
/********************************************************************************
 * Copyright (c) 2023 STMicroelectronics and others.
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License v. 2.0 which is available at
 * http://www.eclipse.org/legal/epl-2.0.
 *
 * This Source Code may also be made available under the following Secondary
 * Licenses when the conditions for such availability set forth in the Eclipse
 * Public License v. 2.0 are satisfied: GNU General Public License, version 2
 * with the GNU Classpath Exception which is available at
 * https://www.gnu.org/software/classpath/license.html.
 *
 * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0

// based on https://github.com/TypeFox/monaco-languageclient/blob/vwj-2.0.1/packages/vscode-ws-jsonrpc/src/socket/reader.ts
/* --------------------------------------------------------------------------------------------
 * Copyright (c) 2018-2022 TypeFox GmbH (http://www.typefox.io). All rights reserved.
 * Licensed under the MIT License. See License.txt in the project root for license information.
 * ------------------------------------------------------------------------------------------ */
 ```
Or the way we did it in the Theia-integration:

```ts
/********************************************************************************
 * Copyright (c) 2017-2018 TypeFox and others.
 * Modifications: (c) 2023 EclipseSource and others.
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License v. 2.0 which is available at
 * http://www.eclipse.org/legal/epl-2.0.
 *
 * This Source Code may also be made available under the following Secondary
 * Licenses when the conditions for such availability set forth in the Eclipse
 * Public License v. 2.0 are satisfied: GNU General Public License, version 2
 * with the GNU Classpath Exception which is available at
 * https://www.gnu.org/software/classpath/license.html.
 *
 * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
// based on: https://github.com/eclipse-sprotty/sprotty-theia/blob/v0.12.0/src/theia/diagram-commands.ts
```

packages/protocol/src/utils/contribution-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+        return (
+            typeof arg !== 'function' &&
+            // https://github.com/eclipse-theia/theia/issues/3204#issue-371029654
+            // In InversifyJS `4.14.0` containers no longer have a property `guid`.
+            ('guid' in arg || 'parent' in arg)
+        );
+    }
+}
+
+export function bindContributionProvider(bindable: Bindable, id: symbol): void {
```
tortmayr, line 73: Should reuse our already existing `BindingContext` types

```ts
export function bindContributionProvider(context: Pick<BindingContext, 'bind'> | interfaces.Bind, id: symbol): void {
    const bind = typeof context === 'object' ? context.bind.bind(context) : context;
    bind(ContributionProvider)
        .toDynamicValue(ctx => new ContainerBasedContributionProvider(id, ctx.container))
        .inSingletonScope()
        .whenTargetNamed(id);
}
```

tortmayr (changes requested):

packages/protocol/src/utils/contribution-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+ *******************************************************************************/
+// eslint-disable-next-line max-len
+// from https://github.com/eclipse-theia/theia/blob/9ff0cedff1d591b0eb4be97a05f6d992789d0a24/packages/core/src/common/contribution-provider.ts
+
+import { interfaces } from 'inversify';
+
+export const ContributionProvider = Symbol('ContributionProvider');
```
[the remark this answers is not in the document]
martin-fleck-at: Sound like a good plan to me! I added this in a new commit.

packages/protocol/src/utils/contribution-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    constructor(
+        protected readonly serviceIdentifier: interfaces.ServiceIdentifier<T>,
+        protected readonly container: interfaces.Container
+    ) {}
+
+    getContributions(recursive?: boolean): T[] {
+        if (this.services === undefined) {
+            const currentServices: T[] = [];
+            let currentContainer: interfaces.Container | null = this.container;
```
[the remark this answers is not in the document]
martin-fleck-at: Good idea!

packages/protocol/src/utils/contribution-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+            // eslint-disable-next-line no-null/no-null
+            while (currentContainer !== null) {
```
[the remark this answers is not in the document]
martin-fleck-at: Done

packages/protocol/src/utils/contribution-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+                if (currentContainer.isBound(this.serviceIdentifier)) {
+                    try {
+                        currentServices.push(...currentContainer.getAll(this.serviceIdentifier));
+                    } catch (error) {
+                        console.error(error);
+                    }
+                }
+                // eslint-disable-next-line no-null/no-null
+                currentContainer = recursive === true ? currentContainer.parent : null;
```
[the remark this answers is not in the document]
martin-fleck-at: Done

packages/protocol/src/utils/contribution-provider.ts
martin-fleck-at: Done

packages/protocol/src/utils/contribution-provider.ts
martin-fleck-at: Done, I chose the Theia-integration way.

packages/protocol/src/utils/contribution-provider.ts
martin-fleck-at: Done

packages/client/src/base/feedback/feedback-command.ts
```diff
@@ -13,11 +13,16 @@
...
 
-export abstract class FeedbackCommand extends Command {
-    // used by the `FeedbackAwareUpdateModelCommand`
-    readonly priority: number = 0;
+export abstract class FeedbackCommand extends Command implements Ranked {
+    /** @deprecated Use rank instead. Please note that a lower rank implies higher priority, so the order is reversed. */
+    readonly priority?: number = 0;
+
+    // backwards compatibility: convert any existing priority to an equivalent rank
+    readonly rank: number = this.priority ? -this.priority : Ranked.DEFAULT_RANK;
```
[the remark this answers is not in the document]
martin-fleck-at: I mean we already consider this in `getFeedbackRank` so I'm not sure I also want to consider it in `Ranked.getRank()`. What would be the benefit of that?

packages/glsp-sprotty/src/types.ts
```diff
@@ -13,7 +13,10 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
+import type { ContributionProvider as IContributionProvider } from '@eclipse-glsp/protocol';
+import { ContributionProvider } from '@eclipse-glsp/protocol';
 import { TYPES as SprottyTYPES } from 'sprotty';
+
```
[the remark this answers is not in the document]
martin-fleck-at: I pushed a commit that should fix that.

packages/client/src/base/feedback/feedback-command.ts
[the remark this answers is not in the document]
tortmayr: Ah, I didn't see that. My bad.

packages/glsp-sprotty/src/types.ts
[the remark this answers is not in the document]
tortmayr: Not sure if partial export is the best solution here. I mean what's the point of having this in the shared protocol if its not exported?
Anyways, I quickly double checked this and technically we would have only two use cases for using the protocol directly wihout the glsp-client:
- usage in node-server
- usage in vscode-integration (host extension)

The node-server already seems to have another mechanism in place to decouple potential injection-cycles. Using factories for constructing multibindings at runtime (https://github.com/eclipse-glsp/glsp-server-node/blob/4e5a7a9d306c282f5566497be5386a022f984afb/packages/server/src/common/actions/action-handler-registry.ts#L57)
The vscode-integration doesn't use inversify at all in the host-extension.

So maybe it would just be easier to move the `contribution-provider` directly into the client?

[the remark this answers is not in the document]
martin-fleck-at: Yeah, if we ever need it in the node server, we either make a copy since it is basically just a single, very stable class or figure out another way. I moved the code into the client, thank you!

PR #324, opened 2024-03-04
Ensure hidden bounds calculation works in all browser

martin-fleck-at:
#### What it does

- Do not use 'display: none' for hidden getBBox() calculation

- Only re-calculate the bounds for certain elements -- Extend LocalRequestBoundsAction with element IDs -- Restrict calculation to given set of elements if IDs are present

- Minor: Fix re-export for decorationModule

Fixes https://github.com/eclipse-glsp/glsp/issues/1258

#### How to test

- Resize elements or create new elements in the workflow example using the ghost elements (default).
- Check that we have the same visual feedback in both Chrome, Firefox or any other browser you are using

#### Changelog

### DIFF
--- a/packages/client/css/glsp-sprotty.css
+++ b/packages/client/css/glsp-sprotty.css
@@ -73,11 +73,6 @@
     fill: #1d80d1;
 }
 
-.sprotty-hidden .sprotty-resize-handle {
-    /** resize handles should not be considered as part of the elements bounds */
-    display: none;
-}
-
 .sprotty-edge {
     fill: none;
     stroke-width: 1.5px;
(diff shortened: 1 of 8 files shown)


packages/client/css/glsp-sprotty.css
```diff
@@ -73,11 +73,6 @@
     fill: #1d80d1;
 }
```
tortmayr, line 75: Quick question: How do we now make sure that resize handles are not considered for bounds computation?

tortmayr (approved):
Looks good to me and everything now seems to work as expected in Chrome and Firefox.

packages/client/css/glsp-sprotty.css
martin-fleck-at: And a very good question at that! Unfortunately, I simply couldn't reproduce the original issue anymore. It seems that the `getBBox` simply returns the correct bounds now on all browsers for all elements.

PR #325, opened 2024-03-04
Ensure we do not return an invalid anchor point during edge routing

martin-fleck-at:
#### What it does

Fixes https://github.com/eclipse-glsp/glsp/issues/1270

Basically the problem is that the `DiamondAnchor` may return an invalid point during its calculation if you give it an element which has as center the given refPoint, resulting in a 0-length PointToPointLine and ultimately to a division by zero.

#### How to test

- See recording in https://github.com/eclipse-glsp/glsp/issues/1270

#### Changelog


packages/client/src/features/routing/glsp-polyline-edge-router.ts
```diff
@@ -0,0 +1,32 @@
...
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+import { GConnectableElement, GParentElement, GRoutableElement, Point, PolylineEdgeRouter } from '@eclipse-glsp/sprotty';
+import { injectable } from 'inversify';
+
+@injectable()
+export class GLSPPolylineEdgeRouter extends PolylineEdgeRouter {
```
tortmayr, line 20: If I recall correctly we decided in a previous discussion regarding the naming of the feature modules to not use the GLSP prefix if possible and rather rexport the parent sprotty class with a sprotty prefix.

So we should probably change this to 
```suggestion
export class PolylineEdgeRouter extends SprottyPolylineEdgeRouter {
```
and add the corresponding rexport to the glsp-sprotty package

martin-fleck-at: Good point, I pushed an update that properly deals with the sprotty-fixed versions and the re-export!

tortmayr (approved):
Thanks for the update. LGTM!

PR #329, opened 2024-03-18
Ensure that markers are properly reinitialized on update

tortmayr:
#### What it does
Ensures that validation markers are properly reinitialized on model update.
Fixes https://github.com/eclipse-glsp/glsp/issues/1290

The previous approach only cleared markers for with the same `reason` as the new `SetMarkersAction`. 
This lead to double rendering of marker with other reasons.

Details:

Assume that we have markers applied to the model triggered by `batch` validation and now we trigger the live validation 
due to a model update.

- When the client processes the update it reapplies the markers from the previous batch and live validation
- Then the `SetMarkersAction` is processe which in terms calls `registerValidationFeedback`.
 Register validation feedback now only clears the validation markers for the `live` reason, but then applies
 both the new live markers and the old batch markers again => double rendering of batch markers.

With the new approach we clear all markers when `regiisterValidationFeedbackAction` and remove the old feedback.
Then we update the markers for the current reason in the internal map and apply a new Feedback action which applies all markers. (new markers for the current reason, existing markers for other reasons)

#### How to test

The test scenario described in eclipse-glsp/glsp/issues/1290 should now work as expected.

#### Changelog

- [] This PR introduces a breaking change (if yes, provide more details below for the changelog and the migration guide)

### DIFF
--- a/packages/client/src/features/validation/validate.ts
+++ b/packages/client/src/features/validation/validate.ts
@@ -13,26 +13,26 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
-import { inject, injectable, optional } from 'inversify';
 import {
     Action,
     CommandExecutionContext,
     CommandReturn,
     DeleteMarkersAction,
+    GParentElement,
     IActionDispatcher,
     IActionHandler,
     ICommand,
     Marker,
-    GParentElement,
     SetMarkersAction,
     TYPES,
     hasArrayProp
 } from '@eclipse-glsp/sprotty';
+import { inject, injectable, optional } from 'inversify';
 import { EditorContextService } from '../../base/editor-context-service';
 import { IFeedbackActionDispatcher, IFeedbackEmitter } from '../../base/feedback/feedback-action-dispatcher';
 import { FeedbackCommand } from '../../base/feedback/feedback-command';
 import { removeCssClasses } from '../../utils/gmodel-util';
-import { GIssueMarker, createGIssue, getOrCreateGIssueMarker, getGIssueMarker, getSeverity } from './issue-marker';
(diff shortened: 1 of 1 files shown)


planger:
I have to admit I didn't have the time yet to test this, but just from a brief look at the code I already wanted to ask quickly:
Doesn't this now remove all markers independent of their original reason?
The actual idea was to always only remove the markers that correspond to the reason of the validation. The thinking of why this it was implemented this way has been discussed [here](https://github.com/eclipse-glsp/glsp/discussions/1281#discussioncomment-8742302).
Is the actual issue not that the workflow validator returns overlapping markers across reasons?
Thank you!

tortmayr:
> I have to admit I didn't have the time yet to test this, but just from a brief look at the code I already wanted to ask quickly: Doesn't this now remove all markers independent of their original reason? The actual idea was to always only remove the markers that correspond to the reason of the validation. The thinking of why this it was implemented this way has been discussed [here](https://github.com/eclipse-glsp/glsp/discussions/1281#discussioncomment-8742302). Is the actual issue not that the workflow validator returns overlapping markers across reasons? Thank you!

Not the implementation of the workflow validator is correct.
The problem is that existing markers are applied twice.
Once when the model updates (as part of the feedback actions) and a second time
when the feedback is [re-registered](https://github.com/eclipse-glsp/glsp-client/blob/5e435038399d7acf5408a64348e212b8393ea205/packages/client/src/features/validation/validate.ts#L65).

By clearing all markers before the re-registration we avoid this double application.
Here is a quick demo to show that marker removal still works as expected:

https://github.com/eclipse-glsp/glsp-client/assets/2311075/afa5a9c5-9b95-41fb-b769-8254ef2c1d53

tortmayr:
@planger  I have pushed an update using separate feedback emitters per reason.
This should make the code clearer and also avoids unnecessary removal and reapplying of markers.

planger (approved):
Thank you for clarifying this and even further improving the code. Looks good to me! :+1:

PR #330, opened 2024-03-22
GH-1225: Avoid circular dependency issues in multi services

tortmayr:
#### What it does

- Introduce `LazyInjector` that allow deferred retrial of services
 from the container
- Introduce a new `preDiagramLoad` hook for `IDiagramStartup` this hook is calle directly after
 `DiagramLoader.load` is invoked.  This is the central entry point to initialize/activate lazily injected services
- Migrate retrieval of multiInjected services to use the `LazyInjector` instead
   - Defer registration/initialization of multiInjected services from ´postConstruct` to `preDiagramLoad` (where possible)
- Migrate `IGModelListener` API from commandstack to the `EditorContextService`. This serves two purposes: 
  - The `EditorContextService` becomes the central component for managing model related context information
  -  The `Commandstack` is removed from the dependency chain for editor context listeners. This should  
     avoid circular dependency issues related to injecting the action provider. 
- Restructure `di-util` of protocol package into `di` sub directory and add tests

Note that base sprotty still uses multi inject for some services (ViewRegistration, ModelElementRegistrations, IVnode post processors etc.)
Currently, these bindings don't seem to cause any dependency issues. If we encounter additional issues in the future we might consider fully replacing all sprotty multiinjections with contribution providers

Also
- Add `inversify` peerDependency to client and glsp-sprotty package. Adding this dependency explicitly allows the TS-LSP to properly resolve inversify for auto-importing and quick fixing of imports.
- Add option for advanced inversify logs to standalone example. If the example is opened with
`?inversifyLog=true` extensive infos about the inversify resolution tree will be logged
- Fix visibility of projection bars on hidden div
- Fix test launch configs

Fixes https://github.com/eclipse-glsp/glsp/issues/1225

#### How to test
Application should work as before.
In addition, creating "fat" services that implement a lot of API interfaces should now be possible.
I have added a test commit that contributes the `CircularDepTester` a singeltion services which implements all of how available API hooks and injections  (previously) problematic classes like action dispatcher, editor context and selection service.
This should now work without errors.

#### Changelog

- [x] This PR introduces a breaking change (if yes, provide more details below for the changelog and the migration guide)
 No public API breaks, but some minor breaks in protected methods/fields of default implementations. That might affect adopters that customize those default implementations. We need to discuss how to properly mark such breaks.


packages/glsp-sprotty/src/types.ts
```diff
@@ -48,5 +48,7 @@ export const TYPES = {
     IDiagramOptions: Symbol('IDiagramOptions'),
     IDiagramStartup: Symbol('IDiagramStartup'),
     IToolManager: Symbol('IToolManager'),
-    IContributionProvider: Symbol('IContributionProvider')
+    IContributionProvider: Symbol('IContributionProvider'),
+    IContributionInitializer: Symbol('IContributionInitializer'),
+    IServiceProvider: Symbol('IServiceProvider')
```
martin-fleck-at, line 53: I don't think the two new symbols are used anywhere, probably a leftover of some intermediate changes?
I believe even the `IContributionProvider` could be removed.

packages/client/src/features/navigation/navigation-action-handler.ts
```diff
@@ -38,6 +37,7 @@ import {
     hasObjectProp,
     hasStringProp
 } from '@eclipse-glsp/sprotty';
+import { inject, injectable } from 'inversify';
```
martin-fleck-at, line 40: We should try to avoid non-relevant changes like these.

packages/client/src/features/validation/validate.ts
```diff
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019-2023 EclipseSource and others.
+ * Copyright (c) 2019-2024 EclipseSource and others.
```
martin-fleck-at, line 2: We should try to avoid non-relevant changes like these.

packages/client/src/base/action-handler-registry.ts
```diff
@@ -14,56 +14,57 @@
...
+        this.lazyInjector
+            .getAll<ActionHandlerRegistration>(TYPES.ActionHandlerRegistration)
+            .forEach(registration => this.register(registration.actionKind, registration.factory()));
+        this.lazyInjector
+            .getAll<IActionHandlerInitializer>(TYPES.IActionHandlerInitializer)
+            .forEach(initializer => this.initializeActionHandler(initializer));
+    }
+}
+
+let baseClassDecorated = false;
```
martin-fleck-at, line 61: I've seen this pattern a few times now, is this something that can be added to our DI utilities (e.g., `decorateUnamanged`)? We could keep a map of targets that we already re-decorated with unmanaged. Just because it is basically always the same code with just the target and the parameter indices as arguments.

packages/client/css/glsp-sprotty.css
```diff
@@ -230,6 +230,10 @@
     opacity: 1;
 }
 
+.sprotty.sprotty-hidden .sprotty-projection-bar {
+    visibility: hidden;
```
martin-fleck-at, line 234: What is the reason we need to hide them explicitly?

examples/workflow-standalone/src/di.config.ts
```diff
@@ -39,5 +41,40 @@ export default function createContainer(options: IDiagramOptions): Container {
     bindOrRebind(container, TYPES.ILogger).to(ConsoleLogger).inSingletonScope();
     bindOrRebind(container, TYPES.LogLevel).toConstantValue(LogLevel.warn);
     container.bind(TYPES.IMarqueeBehavior).toConstantValue({ entireEdge: true, entireElement: true });
+    configureInversifyLogger(container);
     return container;
 }
+
+function configureInversifyLogger(container: Container): void {
```
martin-fleck-at, line 48: Should we add that to our DI utils with a boolean flag and some default log options? I believe that might be useful for adopters as well.

packages/client/src/base/tool-manager/tool-manager.ts
```diff
@@ -75,27 +79,23 @@ export interface IToolManager {
...
+    readonly defaultTools: Tool[] = [];
+
+    preLoadDiagram(): MaybePromise<void> {
+        const tools: Tool[] = this.lazyInjector.getAll(TYPES.ITool);
+        const defaultTools: Tool[] = this.lazyInjector.getAll(TYPES.IDefaultTool);
+        this.registerTools(...tools);
+        this.registerDefaultTools(...defaultTools);
         this.enableDefaultTools();
-        this.contextServiceProvider().then(editorContext => {
-            editorContext.onEditModeChanged(change => this.editModeChanged(change.newValue, change.oldValue));
```
martin-fleck-at, line 96: Was this removed by accident? I don't see how our `editModeChanged` method will be triggered now and the ToolManager is not bound as a `IEditModeListener`.

martin-fleck-at (changes requested):
Thank you very much for that very valuable contribution. For sure, we should add something in the CHANGELOG for that as well. But for now, everything seems to work well. I do have some minor questions though so maybe you could look at that, thank you!

packages/client/src/features/validation/validate.ts
tortmayr: This get autofixed if you use the `check:headers` command. Seems like we missed to update the header in https://github.com/eclipse-glsp/glsp-client/pull/329

packages/client/src/base/tool-manager/tool-manager.ts
tortmayr: Good catch. Yes, the tool manager should be bound as an IEditModeListener Instead

packages/client/css/glsp-sprotty.css
tortmayr: The projection bars are still visible in the hidden div which leads to a weird render artifact in the top right corner of the standalone example.
I guess the reason for this, that the "hidden div" is not actually hidden its just a div with 0 width and height.

image

examples/workflow-standalone/src/di.config.ts
tortmayr: I don't think that that is a good idea since the logger functionality is provided by an additional package (inversify-logger-middleware).
If we add this to the di-util we would also need to introduce an additional (peer) dependency to `inversify-logger-middleware`.

I would rather keep this local to our standalone example. If adopters want to integrate similar logging functionality they can refer to the example as a reference.

packages/client/src/base/action-handler-registry.ts
tortmayr: After retesting this, I don't think its necessary at all. Just passing an empty array in the constructor seems to be enough.
Seem like there was another underyling issue when I originally tested this and then introduced the fix. But with the current solution we don't need the post decoration at all.

tortmayr:
@martin-fleck-at Thanks for the review.
I have pushed a change that should address the remaining review comments.

> Was this removed by accident? I don't see how our `editModeChanged` method will be triggered now and the ToolManager is not bound as a `IEditModeListener`.

Good catch. I have added the missing `IEditModeListener` binding and also updated the standalone example. It's no possible to test the readonly edit mode via url parameter (readonly="true")

> I've seen this pattern a few times now, is this something that can be added to our DI utilities (e.g., `decorateUnamanged`)? We could keep a map of targets that we already re-decorated with unmanaged. Just because it is basically always the same code with just the target and the parameter indices as arguments.

I was not happy with the unmananged decoration of the base class in general. This can be  dangerous since the decoration is global and once decorated the base class can no longer be properly injected in other GLSP/sprotty implementations that are loaded in the same application.

So I played around with different potential solutions. It seems like the constructor of the subclass inherits the decorations of the
parent constructor unless explicitly overwritten. 
So this does not work

```ts
@injectable()
export class BaseClass {
 constructor(@multiInject("ServiceId") services:unknown[]){
 }
}
@injectable()
export class SubClass extends BaseClass {

  constructor(){
     super([]);
  }
}
```
Even though the super constructor parameter is never used inversify will still resolve it
on injection  => we end up with the same circular dependency issues

The cleanest solution is to override the implicit injection of the parent constructor with an explicit injection in the sub class:
```ts
@injectable()
export class SubClass extends BaseClass {

  constructor(@optional() services:unknown[]=[]){
     super(services);
  }
}
```
This works as expected and the injected property of the base class is not resolved when injection the subclass.

So I have reworked all services that previously used the `unmanaged` approach to this solution-

packages/client/src/features/validation/validate.ts
martin-fleck-at: Got it, thanks!

examples/workflow-standalone/src/di.config.ts
martin-fleck-at: Ah, I see, I missed the extra dependency. Yeah, then I totally agree that we should not force that one on adopters!

packages/client/src/base/action-handler-registry.ts
martin-fleck-at: I really like the new solution much better, great idea and work!

packages/client/src/base/tool-manager/tool-manager.ts
martin-fleck-at: Tested and works again, thank you!

packages/client/css/glsp-sprotty.css
martin-fleck-at: Ah, good catch! I didn't notice that one before but with your fix it is gone.

martin-fleck-at (approved):
@tortmayr Thank you for finding a more elegant solution to the `unmanaged` problem. I retested your changes and everything works as expected. Great work that will definitely help us solve many problems out of the box and down the line, thank you!

PR #332, opened 2024-04-08
Implemented webworker connection provider

Sakrafux:
Since most of the functionality is already provided by `vscode-jsonrpc`, I merely created a rather simple wrapper in order to conform to the already existing web socket implementation.


Sakrafux:
@tortmayr This PR is necessary for all webworker implementations. 
So far I circumvented any build problem by developing locally, however I would now also need this change published, in order to pull it in my project's github-action.

tortmayr (approved):
LGTM!

PR #334, opened 2024-04-17
GLSP-1320: Update to latest glsp dev config

tortmayr:
#### What it does

- Updates to latest version of `@eclipse-glsp/dev`
- Add `generate:index` utility script
- Regenerate index files
- Fix codeActionsOnSave in vscode settings
- Refactor `upgrade:next` script
- Reformat code base with prettier
- Fix copyright headers (by running a full check with glsp checkHeaders . -t full)

#### Changelog


packages/protocol/src/utils/index.ts
```diff
@@ -16,7 +16,5 @@
 export * from './array-util';
 export * from './disposable';
 export * from './event';
-// we do not export test-util to avoid a dependency on test frameworks such as chai
-// however, adopters can still access the file by accessing it through the complete path
-// export * from './test-util';
+export * from './test-util';
```
martin-fleck-at, line 19: We should create a `.indexignore` file in the utils directory with the following content to avoid the unwanted export:
```
# we do not export test-util to avoid a dependency on test frameworks such as chai
# however, adopters can still use the functionality by importing it through the complete path
test-util.ts
```

packages/protocol/package.json
```diff
@@ -37,6 +37,7 @@
   "scripts": {
     "build": "tsc -b",
     "clean": "rimraf lib *.tsbuildinfo  coverage .nyc_output ",
+    "generate:index": "glsp generateIndex src -f",
```
martin-fleck-at, line 40: I'll just add this comment here but in https://github.com/eclipse-glsp/glsp/issues/1286 it was discussed to remove the export of the inversify utilities (or moving them out) due to the creation of the dependency to `reflect-metadata` that it creates. If that is something that we can simply fix by not re-exporting the utilities, I'd suggest to use the same approach as with the test-utils.

martin-fleck-at (changes requested):
Change looks good to me, thank you Tobias! However, we should definitely avoid to re-export the test-utils again as it creates undesired dependencies.

tortmayr:
Thanks for the review @martin-fleck-at 
I have resolved your review comments and fixed the failing test step

packages/protocol/src/.indexignore
```diff
@@ -0,0 +1,2 @@
+utils/test-util.ts
+di
```
martin-fleck-at, line 2: Nit: Could you please add some comments above each line as to why we are doing this? I don't think it is very obvious, e.g.,:

```suggestion
# do not export test-util in index to avoid a necessary dependency on test frameworks such as chai
# adopters can still use the functionality by importing it through the complete path
utils/test-util.ts
# do not export the DI utilities to avoid a necessary dependency on reflect-metadata
di
```

package.json
```diff
@@ -14,23 +14,28 @@
...
+    "test": "lerna run test --no-bail",
+    "test:ci": "lerna run test:ci --no-bail",
     "test:coverage": "lerna run test:coverage",
     "test:coverage:ci": "yarn glsp coverageReport .",
-    "upgrade:next": "yarn upgrade -p \"@eclipse-glsp.*|sprotty.*\" --next ",
+    "upgrade:next": "glsp updateNext",
     "watch": "concurrently --kill-others -n tsc,standalone -c red,yellow \"tsc -b -w --preserveWatchOutput\" \"yarn -s standalone watch:bundle\""
   },
+  "resolutions": {
+    "snabbdom": "~3.5.1"
```
martin-fleck-at, line 35: Is there a particular reason we need to do this now?

martin-fleck-at (changes requested):
Thank you Tobias, all looks good to me now. I do have a minor request to add some comments above the ignored lines because it is not very obvious why we are doing it. I'd also be interested to know why we need the resolution to snabbdom now.

package.json
martin-fleck-at: Well, I should have read the updated PR description. All good then ;-)

martin-fleck-at (approved):
LGTM!

PR #341, opened 2024-04-23
Provide Geometry utility classes

martin-fleck-at:
#### What it does

- Extend functions for Bounds, Dimensions, and Point
- Provide types and functions for Vector and Movement

Fixes https://github.com/eclipse-glsp/glsp/issues/1322

#### How to test

- Check if tests are running through and all adapted layouting classes still produce the same result.
 
#### Follow-ups
-

#### Changelog

### DIFF
--- a/examples/workflow-glsp/src/workflow-diagram-module.ts
+++ b/examples/workflow-glsp/src/workflow-diagram-module.ts
@@ -31,6 +31,7 @@ import {
     IHelperLineOptions,
     ISnapper,
     LogLevel,
+    Point,
     RectangularNodeView,
     RevealNamedElementActionProvider,
     RoundedCornerNodeView,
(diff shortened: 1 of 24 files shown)


tortmayr (changes requested):
Nice work. Having a central sophisticated base library for geometry related stuff is definitely a huge improvement and should help us with overall stability in the long run.

I know it's cumbersome but I think in order to ensure that all utility methods are working as expected we should properly test them. Since these functions will probably used heavily in adopter projects we have to ensure that they are all working as expected.
Maybe AI can help you reduce the test implementation effort. 😉 

Other than that the code looks good to me and everything seems to be working as expected.

martin-fleck-at:
@tortmayr Good idea! I added some test cases for the functions!

tortmayr (approved):
@martin-fleck-at Thanks for adding the tests. Code looks good to me and everything still works as expected.

PR #344, opened 2024-05-13
Improve handling of change bounds in several areas

martin-fleck-at:
#### What it does

Improve handling of change bounds

- Introduce change bounds manager to centralize bounds-related services
-- Bounds changes through position snapping and movement restriction
-- Validation for size and position of an element
-- Customizable methods for when to use move and resize options

- Introduce change bounds tracker for moves and resizes
-- Tracker calculates move on diagram and calculates move and resizes
-- Tracker supports options on which parts of the process are applied

- Provide moveable wrappers for resize and routing handles

Fixes https://github.com/eclipse-glsp/glsp/issues/1337

- Extend current resize capabilities
-- Introduce mode for symmetric resize
-- Introduce one-dimensional resize on top, right, bottom and left side

Fixes https://github.com/eclipse-glsp/glsp/issues/1338
Fixes https://github.com/eclipse-glsp/glsp/issues/1339

- Fix elements moving during resizing when hitting minimum bounds
-- Store calculated minimum size from layouter in element
-- Adapt resize so we do not produce invalid sized bounds

Fixes https://github.com/eclipse-glsp/glsp/issues/1340

Minor:
- Ensure we get proper cursor feedback when hovering over resize handle
- Add additional convenience functions
- Add origin viewport command for convenience

**_Please note this PR builds upon https://github.com/eclipse-glsp/glsp-client/pull/343 and should only be merged afterwards!_**

#### How to test

- Double check that everything behaves as before (minus the bugs)
- Symmetric resize can be tested using the Ctrl Key
- Additional resize handles can be tested by adding the locations to the ChangeBoundsManager. The corners are kept for backwards-compatibilty.
- General sizing can be tested more easily with the debug bounds module from https://github.com/eclipse-glsp/glsp-client/pull/343

#### Changelog

As soon as we agree on the PR, I'll write a paragraph about the changes ;-)

### DIFF
--- a/packages/client/css/change-bounds.css
+++ b/packages/client/css/change-bounds.css
@@ -2,14 +2,39 @@
     cursor: nw-resize;
 }
 
+.sprotty-resize-handle[data-kind='top'] {
+    cursor: n-resize;
+}
+
 .sprotty-resize-handle[data-kind='top-right'] {
     cursor: ne-resize;
+}
+
+.sprotty-resize-handle[data-kind='right'] {
+    cursor: e-resize;
+}
+
+.sprotty-resize-handle[data-kind='bottom-right'] {
+    cursor: se-resize;
+}
+
+.sprotty-resize-handle[data-kind='bottom'] {
+    cursor: s-resize;
 }
 
 .sprotty-resize-handle[data-kind='bottom-left'] {
     cursor: sw-resize;
 }
 
-.sprotty-resize-handle[data-kind='bottom-right'] {
-    cursor: se-resize;
+.sprotty-resize-handle[data-kind='left'] {
+    cursor: w-resize;
+}
+
+.sprotty-resize-handle.resize-not-allowed {
+    fill: var(--glsp-error-foreground) !important;
+}
+
+.sprotty g .resize-not-allowed > .sprotty-node {
+    stroke: var(--glsp-error-foreground) !important;
+    stroke-width: 1.5px;
 }
--- a/packages/client/css/change-bounds.css
+++ b/packages/client/css/change-bounds.css
@@ -31,10 +31,19 @@
 }
 
 .sprotty-resize-handle.resize-not-allowed {
-    fill: var(--glsp-error-foreground) !important;
+    fill: var(--glsp-error-foreground);
 }
 
 .sprotty g .resize-not-allowed > .sprotty-node {
-    stroke: var(--glsp-error-foreground) !important;
+    stroke: var(--glsp-error-foreground);
     stroke-width: 1.5px;
+}
+
+.move-mode .sprotty-projection-bar,
+.resize-mode .sprotty-projection-bar {
+    /**
+     * We are using mouse events (offsetX, offsetY) in the GLSPMousePositionTracker to calculate the diagram position relative to the parent.
+     * Other elements result in relative coordinates different from the graph and will therefore interfere with the correct position calculation.
+     */
+    pointer-events: none;
 }
--- a/packages/client/css/helper-lines.css
+++ b/packages/client/css/helper-lines.css
@@ -16,8 +16,8 @@
 
 .helper-line {
     pointer-events: none;
-    stroke: #f7086c;
-    stroke-width: 0.5;
+    stroke: #1d80d1;
+    stroke-width: 1;
     opacity: 1;
 }
 
--- a/packages/client/src/utils/gmodel-util.ts
+++ b/packages/client/src/utils/gmodel-util.ts
@@ -14,9 +14,7 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 import {
-    Bounds,
     BoundsAware,
-    Dimension,
     EdgeRouterRegistry,
     ElementAndBounds,
     ElementAndRoutingPoints,
@@ -42,7 +40,6 @@ import {
     isSelected,
     remove
 } from '@eclipse-glsp/sprotty';
-import { GEdge } from '../model';
 
 /**
  * Helper type to represent a filter predicate for {@link GModelElement}s. This is used to retrieve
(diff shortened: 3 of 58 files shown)


packages/client/src/utils/gmodel-util.ts
```diff
@@ -416,3 +463,29 @@ export function isNotDescendantOfAnyElement<T extends GModelElement>(elements: F
...
+ * @returns the bounds of the edge
+ */
+export function getAbsoluteEdgeBounds(edge: GEdge): Bounds {
+    if (!edge.source || !edge.target) {
+        return edge.bounds;
+    }
+    const source = edge.source;
+    const target = edge.target;
+    const edgePoints: Bounds[] = [];
+    edgePoints.push({ ...Dimension.ZERO, ...Bounds.center(source.bounds) });
```
tortmayr, line 480: Am I misunderstanding something, or is this the hard coded assumption that source and target are always anchored in center? Which would not be correct.

packages/protocol/src/utils/type-util.ts
```diff
@@ -59,6 +59,11 @@ export type MaybePromise<T> = T | PromiseLike<T>;
  */
 export type TypeGuard<T> = (element: any) => element is T;
 
+/** Utility function to combine two type guards */
+export function typeGuard<T, G>(one: TypeGuard<T>, other: TypeGuard<G>): TypeGuard<T & G> {
```
tortmayr, line 63: We could allow for potentially merging more than 2 guards if we spreat the other argument

tortmayr (changes requested):
Thanks a lot Martin.
I really like the centralized approach of handling bound changes.
I have retested everything and can confirm that all mentioned bugs are fixed and  (almost) everything works as expected.

There seems to be a bug when resizing elements close to the diagram border when projetion bars are visible:


https://github.com/eclipse-glsp/glsp-client/assets/2311075/d3157686-119a-4938-a8b0-fa775fa5c595

On master the resize works as expected in this case.

Other than that i also noticed that the a11y keyboard resize tool currently does not respect grid size i.e. resizing with +/- does not snap the size to the grid. But we can handle that in a follow-up PR if necessary.

packages/client/src/utils/gmodel-util.ts
```diff
@@ -155,6 +158,50 @@ export function removeCssClasses(root: GModelElement, cssClasses: string[]): voi
     remove(root.cssClasses, ...cssClasses);
 }
 
+/**
+ * Adds a css classs to a set of {@link GModelElement}s.
+ *
+ * @param elements The elements to which the css class should be added.
+ * @param cssClass The css class to add.
+ */
+export function addCssClassToElements(elements: GModelElement[], cssClass: string): void {
```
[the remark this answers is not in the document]
martin-fleck-at: Good catch, I'll clean that up!

packages/client/src/utils/gmodel-util.ts
martin-fleck-at: Very good point! I'm not sure we can provide a generic method without handing in at least the edge registry. Since this is currently not used, I may just skip it. Thanks!

packages/protocol/src/utils/type-util.ts
martin-fleck-at: Hmm... how does it work with the Generics arguments though? Because right now it combines `T` and `G` but if we are using varargs, I don't think we can properly specify the type for the return value. But my Typescript knowledge limit my thinking here ;-)

packages/client/css/change-bounds.css
```diff
@@ -2,14 +2,39 @@
...
+.sprotty-resize-handle[data-kind='bottom-left'] {
+    cursor: sw-resize;
+}
+
+.sprotty-resize-handle[data-kind='left'] {
+    cursor: w-resize;
+}
+
+.sprotty-resize-handle.resize-not-allowed {
+    fill: var(--glsp-error-foreground) !important;
```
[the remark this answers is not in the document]
martin-fleck-at: Yes, I think we can actually skip it in this case as there is no more specific rule as far as I can see, good catch!

packages/protocol/src/utils/type-util.ts
tortmayr: True. I overlooked that somehow. Please ignore the initial comment 😉

martin-fleck-at:
@tortmayr I think I managed to address all the issues, I'd really appreciate it if you could have another, more detailed look at this. Thank you very much!

packages/client/css/helper-lines.css
```diff
@@ -16,8 +16,8 @@
 
 .helper-line {
     pointer-events: none;
-    stroke: red;
-    stroke-width: 1px;
+    stroke: #f7086c;
```
tortmayr, line 19: Should we maybe use a less aggressive color than red for the helper lines?

packages/protocol/src/sprotty-geometry-bounds.ts
```diff
@@ -49,7 +49,7 @@ declare module 'sprotty-protocol/lib/utils/geometry' {
          * @param right right bounds
          * @returns true if the two bounds are equal
          */
-        function equals(left: Bounds, right: Bounds): boolean;
+        function equals(left: Bounds, right: Bounds, eps?: number): boolean;
```
tortmayr, line 52: Missing @param for the eps param

tortmayr (changes requested):
Thanks Martin. All issues have been addressed and everything seems to work as expected now. Nice job!

I encountered one issue with the recently introduced debug module. It also decorates validation markers (and potentially other decorations) which looks a bit confusing. We should probably not decorate decorations at all:

image

packages/protocol/src/sprotty-geometry-bounds.ts
martin-fleck-at: Fixed.

packages/client/css/helper-lines.css
martin-fleck-at: Do you maybe have a suggestion? I tried to choose a softer red, more pink-ish color to make it less look like an error. However, I couldn't quickly find a color that has a good contrast to most other colors and still look very visible on the blue-ish background.

martin-fleck-at: I pushed an update where I set the stroke width back to 1 and use the blue that we also use for the resize handles, please have a look at that.

martin-fleck-at:
@tortmayr Thank you, I pushed an update that should address the remaining issues. Now we just need to agree on a color ;-)

tortmayr (approved):
Thanks Martin.
Everything now works as expected.
I'm happy with the new color of the helper lines 😛

PR #345, opened 2024-05-14
Align sprotty overrides

tortmayr:
#### What it does

- Move glsp-specific overrides of sprotty base classes into glsp-client directory.
- Add rule from ui-extension.css to base css file. This also fixes an issue with the missing css file we currently encounter in downstream builds

Sidenote: The sprotty-glsp package is intended as a small layer to enable a controlled reexport of sprotty provided modules. Apart from the overrides for our custom action definition (action-override), all other customization of sprotty classes should be handled in the glsp-client package.

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #346, opened 2024-05-16
GLSP-1309: Improve error handling of startup hooks

tortmayr:
#### What it does
- Refactor `DiagramLoader` to catch and log errors that occur when invoking startup hooks
- This ensures both: failure of one hook does not affect execution of following hooks and if  a hook fails for any reason we provide the reason in the log output
- Refactor `RestoreViewportHandler` to use the actual dom id of the root element for querying/focusing the graph. This ensures that the focusing also works if the root model element is not of type `graph`.

Fixes https://github.com/eclipse-glsp/glsp/issues/1309

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #347, opened 2024-05-22
Remove eager default prevention from standalone copy paste handlder

tortmayr:
#### What it does

Default event prevention should be handled by the `ICopyPasteHandler` implementation. e.g. The default `ServerCopyPasteHandler` has guard clauses to check wether a copy or paste operation should be executed (diagram must be active and selection >0). It only default prevents if this condition is met. Currently we bypass this special check and always default prevent which causes issues in scenarios where the diagram is nested into a surrounding application frame or is currently not focused (e.g.) hidden.

See also https://github.com/eclipse-glsp/glsp/discussions/1331

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #348, opened 2024-05-24
Improve bounds and insert management and customizability

martin-fleck-at:
#### What it does

- Allow resizable elements to specify resize handle location
- Use correct feedback (move) on ghost element move (not resize)
- Centralize most move/resize feedback in change bounds manager
- Do not draw grid on hidden SVG as we do not have transformation data
- Ensure we also apply move feedback if we do not have ghost element
- Ensure that the container feedback considers the insert position
- Extract container insert management into manager class
- Avoid unnecessary TYPES.* for classes
- Fix typos
- Remove circular dependencies by fixing index imports

Contributed on behalf of Axon Ivy AG

#### How to test

- The insert indicator feedback can be tested by not providing a ghost element in the server.
 
#### Follow-ups

-

#### Changelog


packages/client/src/utils/gmodel-util.ts
```diff
@@ -253,7 +254,9 @@ export type MoveableElement = GModelElement & Locateable;
 
 export interface Resizable extends BoundsAware, Selectable {}
 
-export interface ResizableModelElement extends GParentElement, Resizable {}
+export interface ResizableModelElement extends GParentElement, Resizable {
+    resizeLocations?: ResizeHandleLocation[];
```
tortmayr, line 258: We should mirror this interface to  the server side and the `GShapedElement` should implement it by default.
Otherwise its not possible to define the resize location in the model on the server side.

packages/client/src/views/glsp-svg-views.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    protected getRadius(node: GShapeElement): number {
+        const d = Math.min(node.size.width, node.size.height);
+        return d > 0 ? d / 2 : 0;
+    }
+}
+
+export function hiddenBoundingRect(withBounds: BoundsAware): VNode {
+    // an element with attribute ATTR_BBOX_ELEMENT is used by the hidden bounds updater to determine the bounds
+    return <rect attrs={{ [ATTR_BBOX_ELEMENT]: true }} class-hidden={true} {...Dimension.fromBounds(withBounds.bounds)} />;
```
tortmayr, line 104: Side note: Explicitly setting the hidden class should not be necessary. The bounding rect is only rendered in the hidden viewer anyways => its always hidden by default

packages/glsp-sprotty/src/types.ts
```diff
@@ -48,8 +48,5 @@ export const TYPES = {
     IDiagramOptions: Symbol('IDiagramOptions'),
     IDiagramStartup: Symbol('IDiagramStartup'),
     IToolManager: Symbol('IToolManager'),
-    IDebugManager: Symbol('IDebugManager'),
-    Grid: Symbol('Grid'),
-    IGridManager: Symbol('IGridManager'),
-    IChangeBoundsManager: Symbol('IChangeBoundsManager')
+    Grid: Symbol('Grid')
```
tortmayr, line 51: Shouldn't this be IGrid, for consistency?

packages/client/src/views/glsp-svg-views.tsx
```diff
@@ -0,0 +1,109 @@
...
+    svg
+} from '@eclipse-glsp/sprotty';
+import { injectable } from 'inversify';
+import { VNode } from 'snabbdom';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const JSX = { createElement: svg };
+
+@injectable()
+export class GDiamondNodeView extends ShapeView {
```
tortmayr, line 40: I guess we should deprecate the `DiamondNodeView` and `CircularNodeView`  classes provided by sprotty and mention that the corresponding GView should be used instead.

tortmayr (commented):
Thanks for applying the review comments @martin-fleck-at.

I noticed that there is now a noticeable diff between some nodes already presented in the diagram and newly created ones. (e.g. merge/decision nodes already present cannot be resized in hight, the newly created ones can).

Maybe we should recreate the diagram from scratch to ensure that everything is in sync?

tortmayr:
Also I noticed that the test cases are now failing and pushed a fix for that (imports adustment)

packages/client/src/views/glsp-svg-views.tsx
martin-fleck-at: Good catch! It was from another approach I tried before, thank you for catching that!

packages/client/src/utils/gmodel-util.ts
martin-fleck-at: That is a good point, as soon as this PR is merged, I'll open one for our two server repositories.

packages/client/src/views/glsp-svg-views.tsx
martin-fleck-at: Good point!

packages/glsp-sprotty/src/types.ts
martin-fleck-at: I'd rather not create an interface `IGrid` since it is an alias of `Point` which also does not use the `I`-prefix that is not that common in Typescript anyway. Unless you have very strong feelings about that.

martin-fleck-at:
@tortmayr All comments should be addressed in the new revision.

tortmayr (approved):
LGTM! 👍🏼 Thanks.

PR #350, opened 2024-05-27
Ensure that align feature is exported from sprotty

tortmayr:
Ensure that align feature is exported from sprotty


martin-fleck-at (approved):
LGTM!

PR #352, opened 2024-05-29
Improve some aspects of change bounds, feedback and positioning

martin-fleck-at:
#### What it does

- Ensure GridManager properly tracks state if action is used on startup
- Ensure DebugManager properly tracks state if action is used on startup
- Avoid cursor feedback interfering with edge edit feedback in disposal
- Avoid unnecessary disposal calls by calling mouseUp on mouseMove
- Fix issue with non-resizable elements not being movable
- Fix calculation of position on diagram for nested HTML elements

#### How to test

- GridManager/DebugManager: Send a ShowGridAction, EnableDebugModeAction on startup and see that the feedback is applied even after a server update (e.g., move an element).
- Feedback edge: Edit an edge and hit 'Escape' in the middle, ensure that feedback edge is properly disposed.
- Element Move: Remove the resize feature (or typehint) from an element on the server, check whether in can be moved.
- Diagram Position: Introduce a foreign object within your SVG element or drag/resize towards the projection bars.

#### Changelog

### DIFF
--- a/packages/client/css/change-bounds.css
+++ b/packages/client/css/change-bounds.css
@@ -38,12 +38,3 @@
     stroke: var(--glsp-error-foreground);
     stroke-width: 1.5px;
 }
-
-.move-mode .sprotty-projection-bar,
-.resize-mode .sprotty-projection-bar {
-    /**
-     * We are using mouse events (offsetX, offsetY) in the GLSPMousePositionTracker to calculate the diagram position relative to the parent.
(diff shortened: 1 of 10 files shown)


tortmayr (approved):
👍🏼  Changes look good to me and everything mentioned in "How to test" now seems to work as expected.

Nice catch with the incorrectly calculated position in the mouse position tracker

PR #353, opened 2024-06-04
Minor Improvements

martin-fleck-at:
#### What it does

- Ensure we render/unhide tracked ghost element as soon as possible
- Add ghost element bounds data to arguments sent to server

#### How to test

- Ensure we render/unhide tracked ghost element as soon as possible > only move mouse from tool palette to the graph without moving it on the graph
- Add ghost element bounds data to arguments sent to server > debug and check what is sent to or received on the server side

#### Changelog

### DIFF
--- a/packages/client/src/features/tools/node-creation/node-creation-tool.ts
+++ b/packages/client/src/features/tools/node-creation/node-creation-tool.ts
@@ -15,7 +15,10 @@
  ********************************************************************************/
 import {
     Action,
+    Args,
+    Bounds,
     CreateNodeOperation,
+    Dimension,
     Disposable,
     DisposableCollection,
     GModelElement,
@@ -24,6 +27,7 @@ import {
     Point,
     TYPES,
     TriggerNodeCreationAction,
+    isBoundsAware,
     isCtrlOrCmd,
     isMoveable
 } from '@eclipse-glsp/sprotty';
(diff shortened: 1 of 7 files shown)


packages/client/src/features/tools/node-creation/node-creation-tool.ts
```diff
@@ -158,6 +162,25 @@ export class NodeCreationToolMouseListener extends DragAwareMouseListener {
         return result;
     }
 
+    protected getCreateNodeOperationArgs(insert: TrackedInsert, ctx: GModelElement, event: MouseEvent): Args | undefined {
+        let args = { ...this.triggerAction.args };
+        if (this.tool.changeBoundsManager.snapper) {
+            args = { ...args, snapper: this.tool.changeBoundsManager.snapper.constructor.name };
```
tortmayr, line 168: I don't know the intended purpose for the snapper name argument, but the current approach maybe flawed for a couple of reasons:
- changeBoundsManger.snapper is optional so it can be undefined. In that case wouldn't throw `this.tool.changeBoundsManager.snapper.constructor.name` an error?
- The snapper does not necessarily have a constructor function (i.e. is a class). What would happen if the snapper is e.g. bound to a direct object (via toConstantValue?).
- contructor.name is unreliable. In production builds contructor names get minified. So it can/should not be used a stable identifier

Maybe we should introduce an optional id property in the `ISnapper` interface and use that instead?

tortmayr (commented):
Looks good to me. I only have one question/remark regarding the newly introduced `getOrCreateNodeOperationArgs` method.

packages/client/src/features/tools/node-creation/node-creation-tool.ts
martin-fleck-at: For the optional behavior, that is why we have the `if` before but I agree that getting the name like that may not be the best option. I think I may have overshot with this a bit and I'll remove it again due to the unreliable behavior that you mentioned.

tortmayr (approved):
LGTM!

PR #354, opened 2024-06-05
Fix style handling in SVG exporter

martin-fleck-at:
#### What it does

For some reason our created target elements are not actually instanceof Element even though everything is in their prototype chain. So instead, we now check whether it is an element with tag (HTML, SVG, Math) and whether it has a style property.

#### How to test

- Debug and you'll see that we do not keep the style of SVG elements etc.

#### Changelog


tortmayr (approved):
LGTM!

PR #355, opened 2024-06-06
GLSP-1349 Improve container configuration

tortmayr:
#### What it does

- Add `replace` property to `ModulConfiguration` interface. When resolving feature modules of the replace property, any potential already configured module with the same id will be replaced with the defined module. If there is nothing to replace, the module will be added to the end of the resolved configuration
- Add check for non-unique feature module ids to `resolveContainerConfiguration` function.
- Add option to enable additional debug logging when loading feature modules. This is usefull during development to get insights about why the container modules are not loaded as expected
- Add dedicated `featureIds` to all default modules. If no featureId is provided the number id of the module is reused for creating the featureId. While this works perfectly fine, it makes debugging harder because feature modules are hard to indentify based on their featureid. With this change its now easier to derive the corresponding feature module from the featureid.

Fixes https://github.com/eclipse-glsp/glsp/issues/1349

#### How to test
- Unit tests for the `resolveContainerConfiguration` function are inplace and they are also testing the replace behavior
- To test debugLogging of feature modules you can adapt the `di.config` file of the workflow-standalone example:
```ts
export default function createContainer(options: IDiagramOptions): Container {
    ...
    FeatureModule.DEBUG_LOG_ENABLED=true
    const container = createWorkflowDiagramContainer(
        createDiagramOptionsModule(options),
        {
            add: accessibilityModule,
            remove: toolPaletteModule
        },
        STANDALONE_MODULE_CONFIG
    );
    FeatureModule.DEBUG_LOG_ENABLED=false;
    ...
    }
```

#### Changelog

### DIFF
--- a/packages/client/src/features/export/export-modules.ts
+++ b/packages/client/src/features/export/export-modules.ts
@@ -27,12 +27,15 @@ import {
 import { ExportSvgActionHandler } from './export-svg-action-handler';
 import { GLSPSvgExporter } from './glsp-svg-exporter';
 
-export const exportModule = new FeatureModule((bind, _unbind, isBound) => {
-    const context = { bind, isBound };
-    bindAsService(context, TYPES.HiddenVNodePostprocessor, ExportSvgPostprocessor);
-    configureCommand(context, ExportSvgCommand);
-    bind(TYPES.SvgExporter).to(GLSPSvgExporter).inSingletonScope();
-});
+export const exportModule = new FeatureModule(
+    (bind, _unbind, isBound) => {
+        const context = { bind, isBound };
+        bindAsService(context, TYPES.HiddenVNodePostprocessor, ExportSvgPostprocessor);
+        configureCommand(context, ExportSvgCommand);
+        bind(TYPES.SvgExporter).to(GLSPSvgExporter).inSingletonScope();
+    },
+    { featureId: Symbol('export') }
+);
 
 /**
  * Feature module that is intended for the standalone deployment of GLSP (i.e. plain webapp)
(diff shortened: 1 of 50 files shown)


packages/client/src/features/export/export-modules.ts
```diff
@@ -46,5 +46,5 @@ export const standaloneExportModule = new FeatureModule(
         bind(ExportSvgActionHandler).toSelf().inSingletonScope();
         configureActionHandler(context, ExportSvgAction.KIND, ExportSvgActionHandler);
     },
-    { requires: exportModule }
+    { featureId: Symbol('standaloneExport'), requires: exportModule }
```
martin-fleck-at, line 49: The `exportModule` a few lines up could also profit from an actual Symbol because currently it just gets a number assigned.

packages/protocol/src/di/container-configuration.ts
```diff
@@ -52,8 +55,35 @@ export function resolveContainerConfiguration(...containerConfigurations: Contai
             if (config.add) {
                 distinctAdd(modules, ...asArray(config.add));
             }
+            if (config.replace) {
+                asArray(config.replace).forEach(replace => {
+                    const existingIndex = modules.findIndex(m => m instanceof FeatureModule && m.featureId === replace.featureId);
+                    if (existingIndex >= 0) {
+                        modules[existingIndex] = replace;
+                    } else {
+                        distinctAdd(modules, replace);
```
martin-fleck-at, line 64: I know you defined this as the intended behavior in the function comments but I feel that the `replace` behavior should still log a warning if it does not actually replace another module but defaults to the adding behavior. What do you think?

martin-fleck-at (changes requested):
Change looks good to me and I really like the semantics that `replace` brings with it. Just two minor comments inline.

martin-fleck-at (approved):
LGTM!

PR #356, opened 2024-06-09
Minor cleanup & deprecations

tortmayr:
#### What it does

- Adjust name of `SResizeHandle` and `SResizeHandleview` to the gmodel namespace
and reexport under the old name as deprecated for backwards compatibility.

- Ensure that all diagram modules are constructed as `FeatureModules` instead of plain `ContainerModules`

- Rename modules that still have the glsp prefix and also reexport under old deprecated name

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #357, opened 2024-06-12
Update to Sprotty 1.2.0

tortmayr:
#### What it does
Mostly non-breaking changes. Some definitions have moved to the sprotty-protocol package. However, we can mitigate all potential breaks with our @eclipse-glsp/sprotty reexport.

Fixes https://github.com/eclipse-glsp/glsp/issues/1253

#### Changelog

-   [x ] This PR should be mentioned in the changelog

### DIFF
--- a/packages/protocol/src/re-exports.ts
+++ b/packages/protocol/src/re-exports.ts
@@ -13,7 +13,45 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
-export { Viewport } from 'sprotty-protocol/lib/model';
+export {
+    Alignable,
+    EdgeLayoutable,
+    EdgePlacement,
+    EdgeSide,
+    Expandable,
+    Fadeable,
+    SButton as GButtonSchema,
+    SCompartment as GCompartmentSchema,
+    SEdge as GEdgeSchema,
+    ForeignObjectElement as GForeignObjectElementSchema,
+    SGraph as GGraphSchema,
+    HtmlRoot as GHtmlRootSchema,
+    SIssue as GIssue,
+    SIssueMarker as GIssueMarkerSchema,
+    SIssueSeverity as GIssueSeverity,
+    SLabel as GLabelSchema,
+    SModelElement as GModelElementSchema,
+    SModelRoot as GModelRootSchema,
+    SNode as GNodeSchema,
+    SPort as GPortSchema,
+    PreRenderedElement as GPreRenderedElementSchema,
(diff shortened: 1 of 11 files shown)


packages/protocol/src/re-exports.ts
```diff
@@ -13,7 +13,45 @@
...
+    ModelLayoutOptions,
+    Projectable,
+    Scrollable,
+    Selectable,
+    VAlignment,
+    Viewport,
+    Zoomable,
+    isScrollable,
+    isZoomable
+} from 'sprotty-protocol/lib/model';
```
martin-fleck-at, line 54: Are we missing some things from this commit: https://github.com/eclipse-sprotty/sprotty/commit/3a3a0a0ec9a6d46cbafc1d632751d922fd67a634#diff-e79f06c7d954f3bab90fefa6a847d39bacb660d538d281b0703a58fd6c092b8d?

I'm just wondering because sprotty seems to have made a split between the external `BoundsAware` (position and size property) and the internal `InternalBoundsAware` (with just a bounds property). Similar "issues" with `LayoutableChild` or `ModelLayoutOptions`. In the client package we can now choose to import from `@eclipse-glsp/sprotty` or `sprotty-protocol` and the latter results in a warning that we should use the former but they are not actually interchangeable. I wonder how we should best proceed with this but we definitely should also export the sprotty-protocol interfaces as well somehow, what do you think?

martin-fleck-at (changes requested):
Thank you very much Tobi for the update! I had a look and everything seems to be functioning as expected! I do have a few comments though that I'd like to get resolved before we merge this:

- Sprotty seems to have introduced a mechanism to allow overriding registry entries (https://github.com/eclipse-sprotty/sprotty/commit/d5e7ff5328e62313293035df85e0ba9852b3f97c). I know we have our own registry-subclasses, e.g., `GModelRegistry` to allow overriding and only log a message instead of throwing an error. However, with that new mechanism, maybe we should re-think that or at least point to the override option and function that are now available. What do you think?
- Other comment inline ;-)

packages/protocol/src/re-exports.ts
tortmayr: In GLSP we only use the client side model of sprotty (what sprotty calls internal model). If there are differences between sprotty's interfaces in the internal model and their transfer model (SModel resp. GModelSchema in GLSP) we have to make sure that we always reexport the corresponding internal interface (since that is the one we are actually using).

The name clash is unfortunate but their is no easy way resolve this.

tortmayr:
Thanks for the review Martin, I have addressed the comment regarding the model registry.

martin-fleck-at (approved):
LGTM!

PR #359, opened 2024-06-14
Enhance Grid functionality and customizability

martin-fleck-at:
#### What it does

- Ensure grid is applied on normal graph view without projections
- Allow grid color to be specified through CSS
- Add more properties to the graph for re-use

#### How to test

- Test graph view and graph projection view and try to change the grid color through CSS

#### Changelog


tortmayr (approved):
LGTM

PR #362, opened 2024-06-18
GLSP-1347: Fix autocomplete widget

tortmayr:
#### What it does

- Update `onSelect` method of autocomplete Widget to dispatch a `input` event instead of a `keyup` event. (Since autocompleter 8.x the suggestion update uses `input` instead of `keyup`
- Enable task editor in standalone example by adding an explicit standaloneTaskEditorModule. Fixes https://github.com/eclipse-glsp/glsp/issues/1347

In addition:
- Update dependencies and align with sprotty
- Use fixed version of vscode-jsonrpc to avoid conflicts with Theia
- Restore original container configuration order in `workflow-diagram-module` (was changed on accident in https://github.com/eclipse-glsp/glsp-client/pull/355)

#### How to test

- Trigger task editing mode by selectin a task then press Ctrl+F2
- Verify that task editing works as expected. 
  -  Applying suggestions + validation 
#### Changelog

### DIFF
--- a/examples/workflow-glsp/src/direct-task-editing/task-editor-module.ts
+++ b/examples/workflow-glsp/src/direct-task-editing/task-editor-module.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2024 EclipseSource and others.
+ * Copyright (c) 2019-2024 EclipseSource and others.
  *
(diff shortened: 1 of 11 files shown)


martin-fleck-at (approved):
Great fix, everything looks good and works for me, thank you Tobias!

PR #364, opened 2024-06-20
View improvements

tortmayr:
#### What it does
- Refactor grind background image rendering to a css only approach
- Use @jsx annotion for view files Use @jsx annotation instead of const JSX definition for view files. Adapt eslint config to ignore  unused svg and html imports to avoid explict deactivation of the rule in each view file. Reason:
  - Sprotty base views also use annotations
   - Annotations are more robust. E.g. mocha tests with ts-node fail with an unused local error   when using JSX const

- Add tests for `initializeDiagramContainer`
- Cleanup GLSPProjectionView overrides

#### Changelog

### DIFF
--- a/packages/client/src/views/glsp-projection-view.tsx
+++ b/packages/client/src/views/glsp-projection-view.tsx
@@ -70,7 +70,7 @@ export class GLSPProjectionView extends ProjectedViewportView {
     }
 
     protected getGridStyle(viewport: Readonly<SGraphImpl>, context: RenderingContext): GridStyle {
-        if (context.targetKind === 'hidden' || !this.gridManager?.isGridVisible) {
+        if (context.targetKind === 'hidden' || !this.gridManager) {
             return {};
         }
         const bounds = this.getBackgroundBounds(viewport, context, this.gridManager);
--- a/packages/client/src/views/glsp-projection-view.tsx
+++ b/packages/client/src/views/glsp-projection-view.tsx
@@ -13,7 +13,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
-
+/** @jsx html */
 import {
     Bounds,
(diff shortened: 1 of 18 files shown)


packages/client/src/views/glsp-projection-view.tsx
```diff
@@ -82,11 +79,7 @@ export class GLSPProjectionView extends ProjectedViewportView {
             [GridProperty.GRID_BACKGROUND_Y]: bounds.y + 'px',
             [GridProperty.GRID_BACKGROUND_WIDTH]: bounds.width + 'px',
             [GridProperty.GRID_BACKGROUND_HEIGHT]: bounds.height + 'px',
-            [GridProperty.GRID_BACKGROUND_ZOOM]: viewport.zoom + '',
-            [GridProperty.GRID_BACKGROUND_IMAGE]: this.getBackgroundImage(viewport, context, this.gridManager),
-            backgroundPosition: `${bounds.x}px ${bounds.y}px`,
-            backgroundSize: `${bounds.width}px ${bounds.height}px`
-            // we do not set the background-image directly in the style object, because we want to toggle it on and off via CSS
+            [GridProperty.GRID_BACKGROUND_ZOOM]: viewport.zoom + ''
```
martin-fleck-at, line 82: Question: Do you think it would be worth setting the variables in any case so they could be re-used even if the grid was not actually visible? I'm just wondering about that. We probably do not need it for the hidden rendering but for a deactivated grid someone might still use something like the zoom.

martin-fleck-at (approved):
Change looks good to me and can be merged as is! Great work, thank you Tobias! I only have a small question regarding the grid properties/variables but that should not prevent any merging.

packages/client/src/views/glsp-projection-view.tsx
tortmayr: Fine with me. I pushed an update

martin-fleck-at (approved):

PR #366, opened 2024-06-21
Minor improvements

tortmayr:
#### What it does

- Update example1.wf
- Remove projection specific css styles for hiding the projection bars on mouse enter/leave over the diagram. This is not a generic feature only certain integration types support it (Theia//VS Code) so it should be handled in the integration specific code base
-  Fix timing issue in `GLSPCommandStack` when notifying `GModelRootListeners` after a model change. Properly wait until the listeners a notified before completing the returned promise. This ensures that all model listeners have received the updated model information before the next action is processed by the dispatcher.
- Introduce `dispatcherAfterNextUpdate` method in `GLSPActionDispatcher` that allows to delay the dispatching of actions until the next model update is processed from the server. (Now actions can be queued from both  client and server)
- Update changelog

#### Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,47 @@
 # Eclipse GLSP Client Changelog
 
-## v2.1.0 - active
+## v2.2.0 - active
+
+### Changes
+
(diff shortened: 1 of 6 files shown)


martin-fleck-at (approved):
Change looks good to me and works well. Not sure how best I can test the listener awaiting besides setting a breakpoint but it all looks stable and good. Thank you Tobias!

PR #369, opened 2024-06-27
Fix RestoreViewPortHandler

tortmayr:
#### What it does

Properly CSS escape the root id before using it as query selector

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #370, opened 2024-07-01
Improve extensibility of creation tools

tortmayr:
#### Changelog

### DIFF
--- a/packages/client/src/features/tools/edge-creation/edge-creation-tool.ts
+++ b/packages/client/src/features/tools/edge-creation/edge-creation-tool.ts
@@ -16,6 +16,7 @@
 import {
     Action,
     AnchorComputerRegistry,
+    Args,
     CreateEdgeOperation,
     GModelElement,
     RequestCheckEdgeAction,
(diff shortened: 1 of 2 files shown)


packages/client/src/features/tools/edge-creation/edge-creation-tool.ts
```diff
@@ -147,6 +148,31 @@ export class EdgeCreationToolMouseListener extends DragAwareMouseListener {
...
+     * The user can then create more elements of the same type without having to re-trigger the tool.
+     * By default, continuous mode is enabled if the user holds the CTRL key.
+     * @param element the current model element
+     * @param event
+     */
+    protected isContinuousMode(element: GModelElement, event: MouseEvent): boolean {
+        return isCtrlOrCmd(event);
+    }
+
+    protected getCreateOperation(element: GModelElement, event: MouseEvent, sourceElementId: string, targetElementId: string): Action {
```
martin-fleck-at, line 163: I think we are missing the call of this method in our `nonDraggingMouseUp` handling.

packages/client/src/features/tools/node-creation/node-creation-tool.ts
```diff
@@ -92,19 +92,21 @@ export class NodeCreationTool extends BaseCreationTool<TriggerNodeCreationAction
     }
 }
 
+export interface ContainerPositioningTool extends PositioningTool {
+    readonly containerManager: ContainerManager;
+}
```
martin-fleck-at, line 97: nit: I know we do not enforce this but I still would like to stay consistent within the file when it comes to empty lines. So could you please add an empty line after this declaration? If not, this PR will still be merged since I know it is just a preference ;-)

martin-fleck-at (changes requested):
Change looks good to me for the most part! Just missing the call of the actual function but if that is fixed, it works as expected. Thank you Tobias!

martin-fleck-at (approved):

PR #372, opened 2024-07-02
Use correct lint:ci command

tortmayr:
#### What it does

A wrong lint:ci script has been pushed on accident. This replaces the script with the correct one so that the Jenkins build can pick up linting issues

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #379, opened 2024-07-15
GLSP-1369: Update example1.wf

tortmayr:
#### What it does

- Update example1.wf file to consistently use easy human readable ids to fix currently failing playwright tests and make testing overall easier
- Update `DeselectKeyListener`. Remove condition to check for active resize handles. Resize handles are now longer reseteted on default tool activatin/focus switch. As a consequence the `DeselectKeyListener` no longer has to consider them and can just always trigger deselection

Part of https://github.com/eclipse-glsp/glsp/issues/1369

#### Changelog

### DIFF
--- a/examples/workflow-standalone/app/example1.wf
+++ b/examples/workflow-standalone/app/example1.wf
@@ -6,7 +6,7 @@
     "y": 0
   },
   "id": "sprotty",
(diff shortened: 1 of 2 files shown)


planger (approved):

PR #383, opened 2024-07-23
Update action workflows

tortmayr:
#### What it does

Update action workflows
#### Changelog

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -22,8 +22,9 @@ jobs:
       GLSP_WEBSOCKET_PATH: 'workflow'
       THEIA_URL: 'http://localhost:3000'
       VSCODE_VSIX_ID: 'eclipse-glsp.workflow-vscode-example'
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -12,11 +12,7 @@
 name: 'CodeQL'
 
 on:
-  push:
-    branches: [master]
-  pull_request:
-    # The branches below must be a subset of the branches above
-    branches: [master]
+  workflow_dispatch:
```
martin-fleck-at, line 15: Are we running the CodeQL not on PRs and master anymore?

martin-fleck-at (approved):
Change looks good to me in general, just not sure why we change the CodeQL workflow.

.github/workflows/codeql-analysis.yml
tortmayr: Fixed

martin-fleck-at (approved):
LGTM!

PR #385, opened 2024-08-14
GLSP-1388 Enhance extensibility of `ContainerManager`

ndoschek:
#### What it does

- Added IContainerManager interface for easier customization
- Consume IContainerManager for dependency injection
- Extract isCreationAllowed submethod for element insertion checks

Resolves https://github.com/eclipse-glsp/glsp/issues/1388

#### How to test

Node creation should work as before, this should not impact the functionality at all.

#### Follow-ups

--

#### Changelog

### DIFF
--- a/packages/client/src/features/tools/node-creation/container-manager.ts
+++ b/packages/client/src/features/tools/node-creation/container-manager.ts
@@ -44,19 +44,34 @@ export interface TrackedInsert {
     options: InsertOptions;
 }
 
+export interface IContainerManager {
+    insert(proxy: GModelElement, location: Point, elementTypeId: string, opts?: Partial<InsertOptions>): TrackedInsert;
+    findContainer(location: Point, ctx: GModelElement, evt?: MouseEvent): ContainerElement | undefined;
+    addInsertFeedback(feedback: FeedbackEmitter, trackedInsert: TrackedInsert, ctx?: GModelElement, event?: MouseEvent): FeedbackEmitter;
+}
+
+/**
+ * The default {@link IContainerManager} implementation.
+ * This class class manages the insertion of elements into containers by validating their positions and types,
+ * providing feedback on the insertion process, and determining the appropriate container based on the location and context.
+ */
 @injectable()
-export class ContainerManager {
+export class ContainerManager implements IContainerManager {
     @inject(ChangeBoundsManager) protected readonly changeBoundsManager: ChangeBoundsManager;
 
     insert(proxy: GModelElement, location: Point, elementTypeId: string, opts?: Partial<InsertOptions>): TrackedInsert {
         const options = { ...DEFAULT_INSERT_OPTIONS, ...opts };
         const container = this.findContainer(location, proxy, opts?.evt);
-        let valid = !container || container.isContainableElement(elementTypeId);
+        let valid = this.isCreationAllowed(container, elementTypeId, opts);
         if (valid && (!container || options.validateLocationInContainer)) {
             // we need to check whether the location is valid either because we do not have a container or the option is set
             valid = opts?.validLocationOverwrite ?? this.changeBoundsManager.hasValidPosition(proxy, location);
         }
         return { elementTypeId, container, location, valid, options };
+    }
+
+    protected isCreationAllowed(container: ContainerElement | undefined, elementTypeId: string, opts?: Partial<InsertOptions>): boolean {
+        return !container || container.isContainableElement(elementTypeId);
     }
 
     findContainer(location: Point, ctx: GModelElement, evt?: MouseEvent): ContainerElement | undefined {
--- a/packages/client/src/features/tools/node-creation/container-manager.ts
+++ b/packages/client/src/features/tools/node-creation/container-manager.ts
@@ -46,6 +46,7 @@ export interface TrackedInsert {
 
 export interface IContainerManager {
     insert(proxy: GModelElement, location: Point, elementTypeId: string, opts?: Partial<InsertOptions>): TrackedInsert;
+    isCreationAllowed(container: ContainerElement | undefined, elementTypeId: string, opts?: Partial<InsertOptions>): boolean;
     findContainer(location: Point, ctx: GModelElement, evt?: MouseEvent): ContainerElement | undefined;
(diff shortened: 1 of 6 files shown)


packages/client/src/features/tools/node-creation/container-manager.ts
```diff
@@ -44,21 +44,36 @@ export interface TrackedInsert {
     options: InsertOptions;
 }
 
+export interface IContainerManager {
```
martin-fleck-at, line 47: Do we expect alternative implementations of this interface? I'm just wondering because for some manager classes we do not do the interface+symbol combination, e.g., `ChangeBoundsManager` and `GridManager` whereas for others we do, e.g., `IHelperLineManager`, `IToolManager`.

I wonder if we have any guidelines for it @tortmayr ?

packages/client/src/features/tools/node-creation/node-creation-module.ts
```diff
@@ -30,7 +30,7 @@ import { InsertIndicatorView } from './node-creation-views';
 export const nodeCreationToolModule = new FeatureModule(
     (bind, unbind, isBound, rebind) => {
         const context = { bind, unbind, isBound, rebind };
-        bind(ContainerManager).toSelf().inSingletonScope();
+        bind(TYPES.IContainerManager).to(ContainerManager).inSingletonScope();
```
martin-fleck-at, line 33:
```suggestion
        bindAsService(context, TYPES.IContainerManager, ContainerManager);
```

This ensures that the `ContainerManager` is also bound to itself and so for any adopters injecting it, the binding will still resolve (backwards compatibility).

packages/client/src/features/tools/node-creation/container-manager.ts
```diff
@@ -44,21 +44,36 @@ export interface TrackedInsert {
...
-        let valid = !container || container.isContainableElement(elementTypeId);
+        let valid = this.isCreationAllowed(container, elementTypeId, opts);
         if (valid && (!container || options.validateLocationInContainer)) {
             // we need to check whether the location is valid either because we do not have a container or the option is set
             valid = opts?.validLocationOverwrite ?? this.changeBoundsManager.hasValidPosition(proxy, location);
         }
         return { elementTypeId, container, location, valid, options };
     }
 
+    protected isCreationAllowed(container: ContainerElement | undefined, elementTypeId: string, opts?: Partial<InsertOptions>): boolean {
```
martin-fleck-at, line 73: This is the core change of this PR, right? By default we delegate here to a method that is provided automatically by the `ApplyTypeHintsCommand` which considers the `containableElementTypeIds` set by the server.

I see how a customization might make sense given that we have access to the insert options here but I wonder in general whether it wouldn't be more useful to simply extend the `isContainableElement` method with those arguments and potentially even allow subclasses (e.g., the CategoryNode) to provide a custom implementation of it and only apply the default implementation from the `ApplyTypeHintsCommand` if no custom implementation is given.

What do you think?

martin-fleck-at (changes requested):
Hi Nina, thank you for the PR. It has opened a few questions for me, so hopefully we can get that sorted quickly. Otherwise, everything works as expected.

packages/client/src/features/tools/node-creation/container-manager.ts
ndoschek: Yes, that's correct, this is the main change to help our use case. In our case, we use a more complex logic that relies on another compatibility service rather than just element types. Since we can't inject this service directly into the model element `isContainableElement` method, we needed to customize it up to this point.

packages/client/src/features/tools/node-creation/container-manager.ts
ndoschek: I extracted the interface to allow for custom `ContainerManager` implementations without having to modify the `NodeCreationTool`.

tortmayr: I don't think we have a guideline for that. We adopted the pattern from sprotty and even there its not consistent and sometime interface+symbol are used for injection, othertimes the class  is injected directly.

I don't have a strong opinion on this but if we go the route of the extracted interface we should make it at least consistent for the other manager classes and also bind them via interface.

packages/client/src/features/tools/node-creation/container-manager.ts
martin-fleck-at: I see, so that is fine by me then. We should just make sure that we use that service everywhere then, e.g., I see a similar usage in `KeyboardPointerPosition` maybe that class should also use the service? Could mean that we need to make `isCreationAllowed` public but I'll leave that up to you to see if that is a good fit.

packages/client/src/features/tools/node-creation/container-manager.ts
martin-fleck-at: Fine with me, we can do it as part of this PR or open another issue to align this more consistently. I don't think we will run into any backwards compatibility issues when we do the binding properly.

ndoschek: Sure, I have created a follow-up issue for the remaining *Manager classes and will tackle this right after this PR then.
See https://github.com/eclipse-glsp/glsp/issues/1393

packages/client/src/features/tools/node-creation/container-manager.ts
ndoschek: Thank you Martin for spotting this! Yes, I agree, that would make sense then to re-use this here.
I'll adapt the code accordingly.
However, the accessibility node creation seems to be broken (also on master). So created a follow-up for this: https://github.com/eclipse-glsp/glsp/issues/1394

packages/client/src/features/tools/node-creation/container-manager.ts
martin-fleck-at: Works for me, thank you very much!

packages/client/src/features/tools/node-creation/container-manager.ts
martin-fleck-at: Thank you for creating the issue.

martin-fleck-at (approved):
Thank you for the quick turnaround! LGTM!

ndoschek:
Thank you for the review @martin-fleck-at!
The Eclipse CI was down today, but the Jenkins build was now completed finally.

PR #387, opened 2024-08-22
Ensure anyone who imports the di functionality gets decorated classes

martin-fleck-at:
#### What it does

Ensures that anyone with that imports '@eclipse-glsp/protocol/lib/di' will get the correctly decorated classes without the need to re-specify the decorations.

#### Changelog


ndoschek (approved):
Thank you @martin-fleck-at, I tested the example with the integrated node server as well as the node glsp-server instance and everything works now as expected :sparkles: Thanks!

PR #388, opened 2024-08-23
GLSP-1393 Align Interface usage across *Manager classes

ndoschek:
#### What it does

- Align interface + symbol usage for *Manager classes
  - `ChangeBoundsManager`, `GridManager`, `DebugManager`
- Use bindAsService to ensure backward compatibility
- Update injects to use the symbol instead of the actual manager class

Resolves https://github.com/eclipse-glsp/glsp/issues/1393

#### How to test

Changing bounds, grid toggling and debug mode toggling should work as before, this should not impact the functionality at all.

#### Follow-ups

--

#### Changelog

### DIFF
--- a/examples/workflow-glsp/src/workflow-startup.ts
+++ b/examples/workflow-glsp/src/workflow-startup.ts
@@ -14,15 +14,15 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import { GridManager, IDiagramStartup } from '@eclipse-glsp/client';
-import { MaybePromise } from '@eclipse-glsp/sprotty';
+import { IDiagramStartup, IGridManager } from '@eclipse-glsp/client';
+import { MaybePromise, TYPES } from '@eclipse-glsp/sprotty';
 import { inject, injectable, optional } from 'inversify';
 
 @injectable()
 export class WorkflowStartup implements IDiagramStartup {
     rank = -1;
 
-    @inject(GridManager) @optional() protected gridManager?: GridManager;
+    @inject(TYPES.IGridManager) @optional() protected gridManager?: IGridManager;
 
     preRequestModel(): MaybePromise<void> {
         this.gridManager?.setGridVisible(true);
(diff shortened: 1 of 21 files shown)


examples/workflow-glsp/src/workflow-startup.ts
```diff
@@ -15,14 +15,14 @@
...
-import { MaybePromise } from '@eclipse-glsp/sprotty';
+import { MaybePromise, TYPES } from '@eclipse-glsp/sprotty';
 import { inject, injectable, optional } from 'inversify';
 
 @injectable()
 export class WorkflowStartup implements IDiagramStartup {
     rank = -1;
 
-    @inject(GridManager) @optional() protected gridManager?: GridManager;
+    @inject(TYPES.IGridManager) @optional() protected gridManager?: GridManager;
```
martin-fleck-at, line 25:
```suggestion
    @inject(TYPES.IGridManager) @optional() protected gridManager?: IGridManager;
```

If we are already introducing interfaces for the manager classes, we should only rely on those interfaces when injecting that class. This is true for all injection points for all manager classes.

martin-fleck-at (changes requested):
Thank you very much Nina! The code looks good to me in general, I just have some worries regarding the injection - which in my opinion should use the interfaces.

@tortmayr Are you happy with the general split between type/interface/class?

tortmayr:
>@tortmayr Are you happy with the general split between type/interface/class?

In general yes. I think this is the best solution to give adopters full flexibility for customizations. We just have to make sure that we keep interface and default implementation in sync if we make future changes. i.e. every public method/field of the default implementation needs to already be added to the interface.

I think this is a pattern that we could extend to our other services as well. (e.g. the selection service).

ndoschek:
Thanks for the feedback! I'll address it in the course of the upcoming week!

ndoschek:
Thank you both for your comments. I’ve updated my commit and created a follow-up issue (GH-1400) to apply the same pattern to other services, as Tobias suggested.

martin-fleck-at (approved):
Thank you, Nina! Everything still looks good to me!

PR #393, opened 2024-10-21
GLSP-1393 Update changelog

ivy-lli:
While updating an existing glsp-client from 2.2.1 to 2.3.0-next.381 I ran into the changes from https://github.com/eclipse-glsp/glsp/issues/1393, which were not documented in the Changelog.

#### What it does

- Add changelog entry for change: https://github.com/eclipse-glsp/glsp-client/pull/388

#### How to test

--

#### Follow-ups

--

#### Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Changes
 
 -   [diagram] Ensure that `GLSPMousePositionTracker` correctly calculates the current position in diagram local coordinates [#391](https://github.com/eclipse-glsp/glsp-client/pull/391)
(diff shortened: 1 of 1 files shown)


tortmayr (approved):
Thanks for this commit. 👍🏼 
In general, its not necessary to update the changelog on/after every PR. We mark changes that are relevant for the log in the PR description and then do a bulk changelog update before a release.
Of course you are free to also provide individual changelog updates but just keep in mind that its not strictly necessary.

ivy-lli:
Good to know 👍

PR #394, opened 2024-10-24
GLSP-1416: Align ActionDispatcher API

tortmayr:
#### What it does

- Rename `action-override` module of @eclipse-glsp/sprotty to `api-override` which  a better fit.
- Add additional API methods of the `GLSPActionDispatcher` directly to the `IActionDispatcher` interface and update related code for consistent usage of the interface. Export new `IActionDispatcher` interface via api-override
- Also add API override for `IVNodePostProcessor`

Also:
- Adjust directs imports of `matchesKeystroke` and `toArray` from sprotty.  Import from @eclipse/glsp-sprotty instead

Fixes https://github.com/eclipse-glsp/glsp/issues/1416

#### How to test

Everything should still work as expected.

#### Changelog

### DIFF
--- a/packages/glsp-sprotty/src/api-override.ts
+++ b/packages/glsp-sprotty/src/api-override.ts
@@ -26,7 +26,7 @@ import {
     IButtonHandler as SIButtonHandler,
     ICommandPaletteActionProvider as SICommandPaletteActionProvider,
     IContextMenuItemProvider as SIContextMenuItemProvider,
-    IVNodePostprocessor as SIVnodePostprocessor,
+    IVNodePostprocessor as SIVNodePostprocessor,
     KeyListener as SKeyListener,
     MouseListener as SMouseListener
 } from 'sprotty';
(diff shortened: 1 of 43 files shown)


packages/glsp-sprotty/src/api-override.ts
```diff
@@ -13,28 +13,32 @@
...
     SButtonImpl as GButton,
     SModelElementImpl as GModelElement,
     SModelRootImpl as GModelRoot,
     ICommand,
+    IActionDispatcher as SIActionDispatcher,
     IActionHandler as SIActionHandler,
     IButtonHandler as SIButtonHandler,
     ICommandPaletteActionProvider as SICommandPaletteActionProvider,
     IContextMenuItemProvider as SIContextMenuItemProvider,
+    IVNodePostprocessor as SIVnodePostprocessor,
```
martin-fleck-at, line 29:
```suggestion
    IVNodePostprocessor as SIVNodePostprocessor,
```

martin-fleck-at (approved):
Great change, thank you Tobias! As far as I can tell it should also perfectly preserve backwards compatibility. I just have a few minor naming/comment suggestions but other than that, the change looks good to me.

PR #395, opened 2024-11-11
GLSP-1421: Properly scope standalone copy-paste listeners

tortmayr:
#### What it does

Ensure that the cut/copy/paste listers for the standalonCopyPasteModule are only invoked if the diagram is currently active (i.e. the active dom element is the diagram svg)

Fixes https://github.com/eclipse-glsp/glsp/issues/1421

#### How to test

Copy & Paste on the diagram should work as expected.
If the diagram is not selected directly e.g. focus is on the tool palette the listener should not trigger.
1. Select and copy a part of the diagram (Ctrl+C)
2. Click a button of the tool palette
3. Try to paste with Ctrl+V
4. Ensure that no elemens are added to the diagram

#### Changelog


planger (approved):

PR #396, opened 2024-11-19
GH-1377: Extract move, zoom, shortcut and resize features from accessibility to own module

haydar-metin:
#### What it does

Works on https://github.com/eclipse-glsp/glsp/issues/1377

#### How to test

Use the arrow keys to move the viewport / elements.
Use +/- to zoom in and out.

#### Changelog

The move, zoom, shortcut and resize features have been moved to their own modules. They are now enabled by default.
- The resize and shortcut help view have a standalone module


haydar-metin:
@tortmayr after some thought, do we even require tools for moving/zooming? Maybe `moveElement` would require it, but moving the viewport and zooming shouldn't need a tool or?

tortmayr:
@Haydar I agree. Currently zooming/scrolling via mouse is provided by the `viewportModule`. This module bind the mouse listeners directly instead of using tools. This means that those listeners are always on (independent of the activation state of the tool manager). Listeners that provide the same functionality via keyboard should probably als be bound directly. 
Otherwise we end up in an inconsistent state. (e.g. during edge creation zooming via mouse would be possible, but zooming via keyboard not (because default tools are disabled at that point).

So we should not use tools for zooming/moving.
For the move-tool its a different story, here a tool definitely make sense but maybe we should integrate this into the change bounds tool instead of having a separate tool just for keyboard movement

Thinking about it maybe it even makes sense to integrate the keyboard functionality for zooming/moving/scrolling directly into the corresponding default module (viewportModule, changeBoundsToolModule) instead of having extra a11y modules for them.
WDYT?

haydar-metin:
@tortmayr sounds good to me! I did the following:

- `MoveElementHandler` (ActionHandler) moved to `bounds`, as we don't have a `changeBounds` container module.
- `MoveElementKeyListener` moved to `change-bounds-tool`
- `MoveViewport*`, `Zoom**` moved to `viewport`

tortmayr:
@haydar-metin  Thanks for providing this change.
If its ok for you I will take a look at this PR after the 2.3 release is completed (planned until the end of next week).
Since this PR only partially addresses  https://github.com/eclipse-glsp/glsp/issues/1377  and also introduces some changes to the protocol API we should probably not include it in the current release.
We can then use the next release cycle to further work on https://github.com/eclipse-glsp/glsp/issues/1377 .

packages/client/src/features/accessibility/view-key-tools/grid-zoom-key-tool.ts
```diff
@@ -34,23 +34,23 @@ import { EditorContextService } from '../../../base/editor-context-service';
...
 import * as messages from '../toast/messages.json';
 import { HideToastAction, ShowToastMessageAction } from '../toast/toast-handler';
 
 /**
  * Zoom viewport and elements when its focused and arrow keys are hit.
  */
 @injectable()
-export class ZoomKeyTool implements Tool {
-    static ID = 'glsp.zoom-key-tool';
+export class GridZoomTool implements Tool {
```
tortmayr, line 48: Maybe we rename this use `GridCellZoomTool` to avoid potential confusion with the `Grid` API used for snapping?

packages/protocol/src/action-protocol/element-move.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import { Action } from './base-protocol';
+
+/**
+ * Action for triggering moving of elements. It is similar to `MoveAction` but is used for moving elements relative
+ * to their current position.
+ * The corresponding namespace declares the action kind as constant and offers helper functions for type guard checks
+ * and creating new `MoveElementRelativeAction`.
+ */
+export interface MoveElementRelativeAction extends Action {
```
tortmayr, line 25: This action should not be part of the client-server protocol. Only a limited subset of client-side only actions are part of the protocol. The `MoveElementRealtiveAction` is defacto a feedback action. Dispatching this from the server-side does not make sense. (Movement changes on the server side should typically be operations)

packages/client/src/features/change-bounds/move-element-key-listener.ts
```diff
@@ -0,0 +1,70 @@
...
+    constructor(
+        protected readonly selectionService: SelectionService,
+        protected readonly changeBoundsManager: IChangeBoundsManager,
+        protected readonly grid: Grid = Grid.DEFAULT
+    ) {
+        super();
+    }
+
+    override keyDown(_element: GModelElement, event: KeyboardEvent): Action[] {
+        const selectedElementIds = this.selectionService.getSelectedElementIDs();
```
tortmayr, line 32: Need to be filtered so that only elements that have are actually moveable (`isMoveable` utility function) are moved.

tortmayr:
I also noticed a minor issue with the `MoveElementsHandler`: When moving to the borders of the diagram the handler adjusts the viewport accordingly. Currently this only works as expected when moving to the left or to the top.
In the other directions the canvas only starts moving once the (0.0) point of the element reaches the canvas limit.
Could you maybe fix this as part of this PR. Should only be a minor change where we also have to consider the elements width/height when moving in positive x/y directions:


https://github.com/user-attachments/assets/39619f0d-c004-4b02-83b2-98b51b261526

tortmayr:
Since we have protocol changes here and potentially more are to come please create a follow-up ticket that we have to update the[ protocol doc on the website](https://eclipse.dev/glsp/documentation/protocol/) once the a18y epic is completed.

I like the viewport adjustsments if we reach the canvas bounds on move. Maybe we should add the same functiallity to the change bounds mouse move tool (follow-up).

packages/client/src/features/accessibility/view-key-tools/grid-zoom-key-tool.ts
```diff
@@ -75,18 +75,15 @@ export class ZoomKeyTool implements Tool {
...
                     getAbsolutePositionByPoint(this.editorContextService.modelRoot, action.options.centerCellPosition)
                 );
-            } else if (KeyboardGridKeyboardEventAction.is(action) && action.options.originId === ZoomKeyTool.ID) {
+            } else if (KeyboardGridKeyboardEventAction.is(action) && action.options.originId === GridZoomTool.ID) {
                 if (matchesKeystroke(action.options.event, 'Minus')) {
-                    viewportAction = this.zoomKeyListener.setNewZoomFactor(
-                        this.editorContextService.modelRoot,
-                        ZoomKeyListener.defaultZoomOutFactor
-                    );
+                    viewportAction = this.zoomKeyListener.setNewZoomFactor(this.editorContextService.modelRoot, ZoomFactor.Default.OUT);
```
tortmayr, line 86: Maybe we should provide the ZoomFactor. in/out properties via a injectable options?
This would make it easier for adopters to customize

packages/client/src/features/viewport/viewport-modules.ts
```diff
@@ -52,6 +56,18 @@ export const viewportModule = new FeatureModule(
...
+        bindAsService(context, TYPES.KeyListener, MoveViewportKeyListener);
+        configureActionHandler(context, MoveViewportAction.KIND, MoveViewportHandler);
+
+        bind(ZoomViewportHandler).toSelf().inSingletonScope();
+        bindAsService(context, TYPES.KeyListener, ZoomViewportKeyListener);
+        configureActionHandler(context, ZoomViewportAction.KIND, ZoomViewportHandler);
+
+        bind(ZoomElementHandler).toSelf().inSingletonScope();
+        bindAsService(context, TYPES.KeyListener, ZoomElementKeyListener);
+        configureActionHandler(context, ZoomElementAction.KIND, ZoomElementHandler);
```
haydar-metin, line 70: Use tools

packages/client/src/base/ui-extension/ui-extension.ts
```diff
@@ -26,6 +26,8 @@ export const CSS_HIDDEN_EXTENSION_CLASS = 'hidden';
  */
 @injectable()
 export abstract class GLSPAbstractUIExtension extends AbstractUIExtension {
+    protected toDisposeOnHide = new DisposableCollection();
```
tortmayr, line 29: Nice addition! 👍🏼

packages/client/src/features/change-bounds/resize/resize-tool.ts
```diff
@@ -0,0 +1,110 @@
...
+import { ResizeElementAction, ResizeType } from './resize-handler';
+
+@injectable()
+export class ResizeKeyListener extends KeyListener {
+    @inject(SelectionService)
+    protected readonly selectionService: SelectionService;
+
+    override keyDown(element: GModelElement, event: KeyboardEvent): Action[] {
+        const actions = [];
+        const selectedElementsIds = this.selectionService.getSelectedElementIDs();
```
tortmayr, line 34: Filter down to resizeable elements

packages/client/src/features/shortcuts/shortcuts-manager.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { Disposable, Emitter, type Event } from '@eclipse-glsp/sprotty';
+import { injectable } from 'inversify';
+import type { ShortcutRegistration } from './shortcuts-model';
```
tortmayr, line 19: No need to have a separate file just for one interface.
Can be moved in here

packages/client/src/features/shortcuts/shortcuts-manager.ts
```diff
@@ -0,0 +1,58 @@
+/********************************************************************************
```
tortmayr, line 1: I think we should move this functionality into the default module (i.e. the base package).
Having it as a dedicated feature module means that adopters could choose to remove/not load this module.
This would break all services that inject the shortcut manager

packages/client/src/features/viewport/viewport-handler.ts
```diff
@@ -134,25 +133,29 @@ export class MoveViewportHandler implements IActionHandler {
...
+export class ZoomHandler implements IActionHandler {
     @inject(EditorContextService)
     protected readonly editorContextService: EditorContextService;
     @inject(TYPES.IActionDispatcher)
     protected readonly actionDispatcher: IActionDispatcher;
 
     handle(action: Action): Action | void {
-        if (ZoomViewportAction.is(action)) {
-            return this.handleZoomViewport(action);
+        if (ZoomAction.is(action)) {
```
tortmayr, line 143: Nit: Unncessary typecheck. If we have only one action that is handled you can define the correct type in the method signature directly

```ts
handle(action: ZoomViewportAction)
````

File: packages/client/src/features/viewport/viewport-tool.ts [hunk 11]
```diff
@@ -0,0 +1,50 @@
...
+ ********************************************************************************/
+
+import { TYPES } from '@eclipse-glsp/sprotty';
+import { inject, injectable } from 'inversify';
+import type { IShortcutManager } from '../shortcuts/shortcuts-manager';
+import { BaseTool } from '../tools/base-tools';
+import { MoveViewportKeyListener, ZoomKeyListener } from './viewport-key-listener';
+
+@injectable()
+export class ViewportTool extends BaseTool {
```
tortmayr, line 24: Maybe rename it to `ViewportKeyTool` and add  short doc comment to mention that this tool only handles keyboard related viewport actions. Mouse actions for zooming/paning are currently bound as direct listeners and therefore alway on.

tortmayr (changes requested):

tortmayr (approved):
Thanks a lot. LGTM! 👍🏼

PR #397, opened 2024-11-20
GLSP-1423 Add MovementBehavior to control invalid multi element movements

ivy-lli:
#### What it does

Fixes: https://github.com/eclipse-glsp/glsp/issues/1423

Per default change the behavior that all selected elements will be reset to it's origin location when one move is invalid. But add an `IMovementBehavior` type where this can be changed to the old style.

_Question:_ 
- Should also the red border feedback be on both elements?
- Better name for the options type / flag on the type?

#### How to test

Add to the `examples/workflow-standalone/src/di.config.ts`
```ts
export default function createContainer(options: IDiagramOptions): Container {
  ...
  container.bind(TYPES.IMovementRestrictor).to(NoOverlapMovementRestrictor);
  container.bind(TYPES.IMovementBehavior).toConstantValue({ allElementsNeedToBeValid: true }); //or false for the old behavior
  return container;
}
```

_New behavior:_
Screen Recording 2024-11-20 at 16 02 56

_Old behavior:_
Screen Recording 2024-11-20 at 15 50 45

#### Changelog

### DIFF
--- a/packages/client/src/features/tools/change-bounds/change-bounds-tool-move-feedback.ts
+++ b/packages/client/src/features/tools/change-bounds/change-bounds-tool-move-feedback.ts
@@ -185,14 +185,14 @@ export class FeedbackMoveMouseListener extends DragAwareMouseListener {
         if (!this.tracker.isTracking()) {
             return [];
         }
-        const elementToMove = this.getElementsToMove(target);
-        if (!this.tool.movementBehavior.allElementsNeedToBeValid) {
+        const elementsToMove = this.getElementsToMove(target);
+        if (!this.tool.movementOptions.allElementsNeedToBeValid) {
             // only reset the move of invalid elements, the others will be handled by the change bounds tool itself
-            elementToMove
+            elementsToMove
                 .filter(element => this.tool.changeBoundsManager.isValid(element))
                 .forEach(element => this.elementId2startPos.delete(element.id));
         } else {
-            if (elementToMove.every(element => this.tool.changeBoundsManager.isValid(element))) {
+            if (elementsToMove.every(element => this.tool.changeBoundsManager.isValid(element))) {
                 // do not reset any element as all are valid
                 this.elementId2startPos.clear();
             }
--- a/packages/client/src/features/tools/change-bounds/change-bounds-tool-move-feedback.ts
+++ b/packages/client/src/features/tools/change-bounds/change-bounds-tool-move-feedback.ts
@@ -185,10 +185,18 @@ export class FeedbackMoveMouseListener extends DragAwareMouseListener {
         if (!this.tracker.isTracking()) {
             return [];
         }
-        // only reset the move of invalid elements, the others will be handled by the change bounds tool itself
-        this.getElementsToMove(target)
-            .filter(element => this.tool.changeBoundsManager.isValid(element))
-            .forEach(element => this.elementId2startPos.delete(element.id));
+        const elementToMove = this.getElementsToMove(target);
+        if (!this.tool.movementBehavior.allElementsNeedToBeValid) {
+            // only reset the move of invalid elements, the others will be handled by the change bounds tool itself
+            elementToMove
+                .filter(element => this.tool.changeBoundsManager.isValid(element))
+                .forEach(element => this.elementId2startPos.delete(element.id));
+        } else {
+            if (elementToMove.find(element => !this.tool.changeBoundsManager.isValid(element)) === undefined) {
+                // do not reset any element as all are valid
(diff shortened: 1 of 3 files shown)


packages/client/src/features/tools/change-bounds/change-bounds-tool-move-feedback.ts
```diff
@@ -185,10 +185,18 @@ export class FeedbackMoveMouseListener extends DragAwareMouseListener {
...
-            .filter(element => this.tool.changeBoundsManager.isValid(element))
-            .forEach(element => this.elementId2startPos.delete(element.id));
+        const elementToMove = this.getElementsToMove(target);
+        if (!this.tool.movementBehavior.allElementsNeedToBeValid) {
+            // only reset the move of invalid elements, the others will be handled by the change bounds tool itself
+            elementToMove
+                .filter(element => this.tool.changeBoundsManager.isValid(element))
+                .forEach(element => this.elementId2startPos.delete(element.id));
+        } else {
+            if (elementToMove.find(element => !this.tool.changeBoundsManager.isValid(element)) === undefined) {
```
martin-fleck-at, line 195:
```suggestion
            if (elementToMove.every(element => this.tool.changeBoundsManager.isValid(element))) {
```

Just a minor nitpick but I think it makes the code more readable.

packages/client/src/features/tools/change-bounds/change-bounds-tool.ts
```diff
@@ -63,6 +63,10 @@ import {
 import { FeedbackMoveMouseListener } from './change-bounds-tool-move-feedback';
 import { ChangeBoundsTracker, TrackedElementResize, TrackedResize } from './change-bounds-tracker';
 
+export interface IMovementBehavior {
```
martin-fleck-at, line 66: I'm a little bit on the fence about whether it should be called `IMovementBehavior` (as we have with the `IMarqueeBehavior`) or `IMovementOptions` (as we have with the `IHelperLineOptions`). I think both are probably fine, just wanted to note that down somewhere. We should probably unify that at some point.

martin-fleck-at (changes requested):
I tested this change and everything works as expected, thank you very much, Lukas! I just have a few minor comments and then we can merge this PR in my opinion.

> * Should also the red border feedback be on both elements?
Good question! I can see arguments for both sides as you may not know which move is the culprit if all elements get the erroneous feedback. So I think leaving it only on the actual invalid move for now is fine. 

> * Better name for the options type / flag on the type?
(see inline comment)

packages/client/src/features/tools/change-bounds/change-bounds-tool.ts
ivy-lli: I named it indeed after the MarqueeBehavior (as I invented this too 🙈), however, I think I like Options more, so I will call it options.

packages/client/src/features/tools/change-bounds/change-bounds-tool-move-feedback.ts
```diff
@@ -185,10 +185,18 @@ export class FeedbackMoveMouseListener extends DragAwareMouseListener {
         if (!this.tracker.isTracking()) {
             return [];
         }
-        // only reset the move of invalid elements, the others will be handled by the change bounds tool itself
-        this.getElementsToMove(target)
-            .filter(element => this.tool.changeBoundsManager.isValid(element))
-            .forEach(element => this.elementId2startPos.delete(element.id));
+        const elementToMove = this.getElementsToMove(target);
```
martin-fleck-at, line 188: nit:
```suggestion
        const elementsToMove = this.getElementsToMove(target);
```

packages/client/src/features/tools/change-bounds/change-bounds-tool.ts
```diff
@@ -84,6 +89,7 @@ export class ChangeBoundsTool extends BaseEditTool {
     @inject(EdgeRouterRegistry) @optional() readonly edgeRouterRegistry?: EdgeRouterRegistry;
     @inject(TYPES.IMovementRestrictor) @optional() readonly movementRestrictor?: IMovementRestrictor;
     @inject(TYPES.IChangeBoundsManager) readonly changeBoundsManager: IChangeBoundsManager;
+    @inject(TYPES.IMovementOptions) @optional() readonly movementBehavior: IMovementOptions = { allElementsNeedToBeValid: true };
```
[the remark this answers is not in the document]
ivy-lli: ah thanks 👍

martin-fleck-at (approved):
Great, thank you very much for the quick turnaround! LGTM!

PR #398, opened 2024-11-21
GLSP-1427: Improve diagram loading and commmand handling

tortmayr:
###### Diagram Loader
- Avoid unnecessary dispatch of empty `SetModelAction`. Await the initialize method of the action dispatcher instead (which already dispatches an empty model internally
- Fix behavior of `postRequestModel` hook to actually work as described in the model. Instead of requesting and awaiting the dispatch of the response `SetModelAction` we now simply dispatch the `RequestModelAction` and continue with the`postRequestModelHook`
- Use promise instead of listener for the model initalization constraint. This way we can await the `postModelInitialization` hook . I.e. the DiagramLoader.load methods actually completes once its completely finished (all hooks have been invoked and resolved)

###### UiExtensions
- Fix action of startup UIExtensions. Instead of directly calling their show-Method, a corresponding `SetUIExtensionVisibilityAction` is dispatched. This ensures that the extension only get activated after the initial diagram (empty-model) is available
- Make `preInitialize´ hook of StatusOverlay async. The status overlay is used to print messages during the `loading` phase. By making it async, the loader waits until the overlay is actually visible => now messages get lost.
- Move tool palette initialization into from `preRequest`to`postRequest` hook. This ensures that the server request for retrieving the items is dispatched after the model request => does not block diagram loading
    - Fix unnecessary reload of the palette body if in read-only mode (palette not visible) 

###### CommandStack
- Override methods that populate/manage the internal undo/redo stack to no-ops Commandstack undo/redo is not supported in GLSP. This means managing the stack is just unnecessary overhead

Contributed on behalf of AxonIvy AG
Fixes https://github.com/eclipse-glsp/glsp/issues/1427
#### Changelog

-   [] This PR introduces a breaking change (if yes, provide more details below for the changelog and the migration guide)

### DIFF
--- a/packages/client/src/base/command-stack.ts
+++ b/packages/client/src/base/command-stack.ts
@@ -14,6 +14,7 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 import {
+    CommandExecutionContext,
     CommandStack,
     Disposable,
     DisposableCollection,
@@ -51,6 +52,10 @@ export class GLSPCommandStack extends CommandStack implements Disposable {
         return this.editorContext.onModelRootChanged;
     }
 
+    /**
+     * Client-side undo/redo is not supported in GLSP. The server is responsible for handling undo/redo requests.
+     * If this method get called it's probably a mistake and a warning is logged
+     */
     override undo(): Promise<GModelRoot> {
         this.logger.warn(
             this,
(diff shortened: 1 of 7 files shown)


martin-fleck-at (approved):
As far as I can tell, everything still works as expected so the change looks good to me, thank you very much, Tobias!

ivy-lli (approved):
Changes looks good, thanks 👍

PR #401, opened 2024-11-25
Various minor improvements

tortmayr:
#### What it does

- Merge *-default.ts implementation files of `FeedbackActionDispatcher` and `HelperLineManager` with the the interface definition files. The split is no longer necessary since we resolved potential circular dependency issues
- Ensure that we consistently use the `LazyInjector` over dedicated async providers and deprecate existing provider injection

#### Changelog


packages/client/src/base/focus/focus-tracker.ts
```diff
@@ -29,7 +30,7 @@ export interface FocusChange {
  * Allows querying of the current focus state and the focused root diagram element and the currently focused element within the diagram.
  */
 @injectable()
-export class FocusTracker implements IActionHandler {
+export class FocusTracker implements IActionHandler, IDiagramStartup {
     protected inActiveCssClass = 'inactive';
     protected _hasFocus = true;
```
martin-fleck-at, line 35: I wonder if that should really be true now that we actually track the focus change in the beginning?

packages/client/src/base/focus/focus-tracker.ts
```diff
@@ -57,12 +58,20 @@ export class FocusTracker implements IActionHandler {
...
+    postModelInitialization(): MaybePromise<void> {
+        this.handleFocusChange();
+    }
+
     handle(action: Action): void | Action | ICommand {
         if (!FocusStateChangedAction.is(action)) {
             return;
         }
 
         this._hasFocus = action.hasFocus;
```
martin-fleck-at, line 70: This should probably be initialized correctly somehow.

martin-fleck-at (changes requested):
Change looks good to me overall but I wonder whether the focus tracking needs to be correctly initialized. Also the PR has merge conflicts. Other than that, I like the cleanup and various improvements, thank you Tobias!

tortmayr:
Thanks for the review Martin. I decided to revert the focus tracker fix for now.
As you said, simply setting the focus to true on init is probably not the right approach and we need to come up with a proper solution. Since this potentially also will affect all integration projects i decided to handle this in a follow-up.
It`s a non-critical release anyways so we can fix this issue later in one of the next release cycles.

martin-fleck-at (approved):
I'm fine with handling the focus tracking in a follow-up issue. Other than that, everything looks good to me, thank you Tobias!

PR #402, opened 2024-12-05
GLSP-1438: Improve GLSPClient default implementations

tortmayr:
#### What it does
- Introduce `onCurrentStateChanged` event for `GLSPClient`
- Ensure that `start` and `initializeServer` behave as expected if they are called  again while the previous promise is still pending.
- Restructure members of the default implementations so that get/setters and their _property are grouped together
- Introduce a dedicated namespace for the `Event` interface with utility functions to
  - listen on a certain event exactly once
   - waitUntil a certain event is fired the next time
- Adapt test cases to verify new behavior

Also:
- Update `GLSPActionDispatcher` to use direct injection for registry instead of async provider

#### Changelog

### DIFF
--- a/packages/protocol/src/utils/event.ts
+++ b/packages/protocol/src/utils/event.ts
@@ -46,6 +46,49 @@ export interface Event<T> extends jsonrpc.Event<T> {
     (listener: (e: T) => unknown, thisArgs?: unknown, disposables?: Disposable[]): Disposable;
 }
 
+export namespace Event {
+    /**
+     * Utility function to register a one-time listener for an event. The listener will be disposed
+     * automatically after the next event is fired.
+     * @param event The event to listen to
+     * @param listener The listener function that will be called when the event happens.
+     * @param thisArgs The 'this' which will be used when calling the event listener.
+     * @param disposables An array to which the {@link Disposable} for removing the listener will be added.
+     * @returns a {@link Disposable} to remove the listener again.
+     */
+    export function once<T>(event: Event<T>, listener: (e: T) => unknown, thisArgs?: unknown, disposables?: Disposable[]): Disposable {
+        const toDispose = event(
+            e => {
+                listener(e);
+                toDispose.dispose();
+            },
+            thisArgs,
+            disposables
+        );
+        return toDispose;
+    }
+
+    /**
+     * Utility function to wait for an event to happen. The function will return a promise that will be resolved
+     * when the event is fired. Optionally a predicate can be provided that will be used to filter the event.
+     * If a predicate is provided, the promise will only be resolved when the predicate returns true.
+     * The underlying listener will be disposed automatically when the promise is resolved.
+     * @param event The event to listen to
+     * @param predicate An optional predicate that will be used to filter the event
+     * @returns a promise that will be resolved when the event is fired (and the optional predicate matches)
+     */
+    export function waitUntil<T>(event: Event<T>, predicate?: (e: T) => boolean): Promise<T> {
+        return new Promise<any>(resolve => {
+            const toDispose = event(e => {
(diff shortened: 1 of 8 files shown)


packages/protocol/src/utils/event.ts
```diff
@@ -46,6 +46,49 @@ export interface Event<T> extends jsonrpc.Event<T> {
...
+    /**
+     * Utility function to register a one-time listener for an event. The listener will be disposed
+     * automatically after the next event is fired.
+     * @param event The event to listen to
+     * @param listener The listener function that will be called when the event happens.
+     * @param thisArgs The 'this' which will be used when calling the event listener.
+     * @param disposables An array to which the {@link Disposable} for removing the listener will be added.
+     * @returns a {@link Disposable} to remove the listener again.
+     */
+    export function once<T>(event: Event<T>, listener: (e: T) => unknown, thisArgs?: unknown, disposables?: Disposable[]): Disposable {
```
martin-fleck-at, line 59: It seems like this function is not actually used anywhere in our code. I don't mind more utility functions, juts wondering about it. Also, isn't this basically the same as the `waitUntil` without a predicate?

packages/protocol/src/utils/event.ts
```diff
@@ -46,6 +46,49 @@ export interface Event<T> extends jsonrpc.Event<T> {
...
+    /**
+     * Utility function to wait for an event to happen. The function will return a promise that will be resolved
+     * when the event is fired. Optionally a predicate can be provided that will be used to filter the event.
+     * If a predicate is provided, the promise will only be resolved when the predicate returns true.
+     * The underlying listener will be disposed automatically when the promise is resolved.
+     * @param event The event to listen to
+     * @param predicate An optional predicate that will be used to filter the event
+     * @returns a promise that will be resolved when the event is fired (and the optional predicate matches)
+     */
+    export function waitUntil<T>(event: Event<T>, predicate?: (e: T) => boolean): Promise<T> {
```
martin-fleck-at, line 80: Should this function also provide this `thisArgs` and `disposables`?

martin-fleck-at (approved):
Change looks good to me and everything still seems to work as expected. I just have a question regarding the new Event utility methods but other than that, everything should be good, thank you Tobias!

packages/protocol/src/utils/event.ts
tortmayr: Yes you are right its currently not in use.
However, I know from other projects (working with theia or vscode events) that I quite often have the use case for a self-disposing one-time listener. So having a utility function for that definitely not hurts.
I have not checked the current code base yet, so maybe we already have some places in the code where we could use this function instead of creating the listener manually.
>Also, isn't this basically the same as the waitUntil without a predicate
No, since `waitUntil` doesn`t not allow you to pass a listener function.
Also calling `waitUntil` in void fashion (i.e without awaiting the resolution) is probably an anitpattern because you end up with a floating promise.

packages/protocol/src/utils/event.ts
tortmayr: `thisArgs´ don't make sense in this case since we don't have pass along an explicit listener function that would use the passed thisArgs.
Allowing the `disposables` array is dangerous because this way we expose a possibility to the user to dispose the underlying listener before the promise resolves => we would end up with a promise that never resolves.

packages/protocol/src/utils/event.ts
martin-fleck-at: Fair enough, just wanted to ask.

packages/protocol/src/utils/event.ts
martin-fleck-at: Alright, fair enough.

PR #404, opened 2024-12-06
Remove workaround in GLPSManhattanEdgeRouter

tortmayr:
#### What it does

Remove the override of `applyInnerHandleMoves` in `GLSPManhattanEdgeRouter`. This workaround was introduced as a fix for https://github.com/eclipse-glsp/glsp-client/pull/200 but has since then also been fixed in base sprotty: https://github.com/eclipse-sprotty/sprotty/issues/310 therefore overriding is no longer necessary.

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #407, opened 2024-12-16
GLSP-1449: Ensure that codebase is compatible with inversify >= 6.1.3

tortmayr:
#### What it does
- Remove usage of not-exported constant in lazy injector
- Fix constructor injection of classes that are overwritten to replace constructor injection with lazy injection. To pass the stricter type checks of newer inversify version we need to inject an empty array here.

This ensures that the code base is compatible with all 6.x versions of inversify and we can hard pin/resolve the version required for newer Theia versions while other integrations don't require any change.

Part of https://github.com/eclipse-glsp/glsp/issues/1449

#### How to test

Add a resolutions block in the root package. json
```ts
"resolutions": { "**/inversify":"^6.1.3"}
```
Build and start the application. Everything should still  work as expected.

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #413, opened 2025-02-19
Remove important from sprotty css

haydar-metin:
#### What it does

Works on https://github.com/eclipse-glsp/glsp/issues/1241

- [x] TODO: REMOVE MOVEMENT RESTRICTION BEFORE MERGE
  - It will fix the failing test 

#### How to test

- Trigger validation
- Move a node to be above another node

#### Changelog

### DIFF
--- a/packages/client/css/glsp-sprotty.css
+++ b/packages/client/css/glsp-sprotty.css
@@ -222,25 +222,25 @@
 }
 
 .sprotty g .movement-not-allowed > .sprotty-node {
-    stroke: var(--glsp-error-foreground) !important;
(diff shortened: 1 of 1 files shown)


tortmayr:
Changes look good to me and for the client/standalone example everything seems to work.
However, I quickly tested the change with the theia-integration an there the movement restriction feedback for routing handles is no longer properly applied:

image
With removing the !important it seems that another rule replaces the fill of the routing handles.
We should probably merge this and then retest the different integrations separately.

tortmayr (approved):

PR #415, opened 2025-02-21
Update default styling for labels

tortmayr:
Set `user-select` to none to avoid blinking cursors and/or other indication for text selection when clicking on labels. (Default document text selection of svg text elements is not working anyways)

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #417, opened 2025-03-17
Include root as a findable container for the container manager

haydar-metin:
#### What it does

Closes https://github.com/eclipse-glsp/glsp/issues/1488

#### How to test

Add the following to the node-server 

```
// examples/workflow-server/src/common/workflow-diagram-configuration.ts
            {
                elementTypeId: DefaultTypes.GRAPH,
                deletable: false,
                reparentable: false,
                repositionable: false,
                resizable: false,
                containableElementTypeIds: [types.MANUAL_TASK]
            },
```

Test it with and without the fix. Without the fix, you can still add anything to the root.

#### Changelog


tortmayr (approved):
LGTM! 👍🏼

PR #419, opened 2025-03-26
Hide sprotty sprotty reconnect handles

tortmayr:
#### What it does
Follow-up for https://github.com/eclipse-glsp/glsp/issues/1490 Ensure that the unwanted sprotty reconnect handles are not rendered.

#### Changelog

### DIFF
--- a/packages/client/src/views/routing-point-handle-view.tsx
+++ b/packages/client/src/views/routing-point-handle-view.tsx
@@ -13,7 +13,6 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
-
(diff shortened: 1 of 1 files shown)


packages/client/src/views/routing-point-handle-view.tsx
```diff
@@ -13,7 +13,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
-
+/** @jsx svg */
 import { SRoutingHandleView, type GRoutingHandle, type RenderingContext, type RoutedPoint } from '@eclipse-glsp/sprotty';
```
haydar-metin, line 17: svg import is missing

haydar-metin (approved):

packages/client/src/views/routing-point-handle-view.tsx
tortmayr: Fixed, (not needed anyways)

PR #428, opened 2025-07-08
GLSP-1544: Localize issue marker severity in popup

tortmayr:
#### What it does

Ensure that the severity displayed in the hover poup of issue markers is properly configurable via the generic i18n messages approach.

Fix https://github.com/eclipse-glsp/glsp/issues/1544

#### Changelog


martin-fleck-at (changes requested):
Code looks good to me but I see that the CI is not happy with the E2E tests. I don't think it is related to your code but could you please check?

PR #432, opened 2025-08-22
Enrich LayoutOperation with canvas bounds and Viewport information

jmenzies12:
https://github.com/eclipse-glsp/glsp/issues/1561

- Add canvasBounds and viewport information as optional parameters on LayoutOperation
- Add RequestLayoutAction and corresponding RequestLayoutActionHandler to allow the client to enrich LayoutOperation before it is send to the server.


martin-fleck-at:
@jmenzies12 Thank you very much! On a first glance, this looks really good. We most likely want to move the `RequestLayoutAction` to the VS Code integration only but keep the adapted `LayoutOperation` in the client and make sure that other calls creating the `LayoutOperation` also provide the necessary information. There are also some minor test cases that need to be fixed but since you said you will not have so much time, I gladly take this over.

PR #438, opened 2025-08-27
GLSP-1531: Introduce viewport change event

tortmayr:
#### What it does

Refactor our approach to handle model changes and introduce dedicated `IModelChangeService`.

- Extends the `ICommandStack` interface by adding an event to react to successfull command execution
- Introduce dedicated `IModelChangeService` to encapsulate model change handling from the `EditorContextService` similar to how we do it for the selection change and the `SelectionService`
- Refactor editor context, to use/forward to the model change service

In addition:
- Clean up disposal behavior of FokusTrack

Fixes eclipse-glsp/glsp/issues/1531

#### How to test

For testing a logger can be configured as event listener e.g:

image

#### Changelog

 Potentially breaking changes for Adopters that extend internal API:
 - GLSPCommandStack: 'notifiyListener' method has been removed
 - EditorContextSerivce: 
    - Remove `_modelRoot` and `onModelRootChangedEmitter` properties. Corresponding getters now forward to the ModelChangeService
   - Remove `notifyModelRootChanged` method (Probably no side effect for adopters. Method was only callable from commandstack anyways)

### DIFF
--- a/packages/client/src/base/command-stack.ts
+++ b/packages/client/src/base/command-stack.ts
@@ -19,6 +19,7 @@ import {
     CommandStack,
     Disposable,
     DisposableCollection,
+    Emitter,
     Event,
     GModelRoot,
     ICommand,
@@ -26,7 +27,6 @@ import {
     LazyInjector
 } from '@eclipse-glsp/sprotty';
 import { inject, injectable, postConstruct, preDestroy } from 'inversify';
-import { Emitter } from 'vscode-jsonrpc';
 import { EditorContextService } from './editor-context-service';
 
 @injectable()
--- a/packages/client/src/base/command-stack.ts
+++ b/packages/client/src/base/command-stack.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019-2024 EclipseSource and others.
+ * Copyright (c) 2019-2025 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -15,24 +15,36 @@
  ********************************************************************************/
 import {
     CommandExecutionContext,
+    CommandExecutionData,
     CommandStack,
     Disposable,
     DisposableCollection,
     Event,
     GModelRoot,
     ICommand,
-    LazyInjector,
-    SetModelCommand,
-    UpdateModelCommand
+    ICommandStack,
+    LazyInjector
 } from '@eclipse-glsp/sprotty';
-import { inject, injectable, preDestroy } from 'inversify';
+import { inject, injectable, postConstruct, preDestroy } from 'inversify';
+import { Emitter } from 'vscode-jsonrpc';
 import { EditorContextService } from './editor-context-service';
 
 @injectable()
-export class GLSPCommandStack extends CommandStack implements Disposable {
+export class GLSPCommandStack extends CommandStack implements ICommandStack, Disposable {
     @inject(LazyInjector)
     protected lazyInjector: LazyInjector;
     protected toDispose = new DisposableCollection();
+
+    protected onCommandExecutedEmitter = new Emitter<CommandExecutionData>();
+    get onCommandExecuted(): Event<CommandExecutionData> {
+        return this.onCommandExecutedEmitter.event;
+    }
+
+    @postConstruct()
+    protected override initialize(): void {
+        super.initialize();
+        this.toDispose.push(this.onCommandExecutedEmitter);
+    }
 
     @preDestroy()
     dispose(): void {
@@ -95,13 +107,7 @@ export class GLSPCommandStack extends CommandStack implements Disposable {
     }
     override async execute(command: ICommand): Promise<GModelRoot> {
         const result = await super.execute(command);
-        if (command instanceof SetModelCommand || command instanceof UpdateModelCommand) {
-            this.notifyListeners(result);
-        }
+        this.onCommandExecutedEmitter.fire({ command, newRoot: result });
         return result;
-    }
-
-    protected notifyListeners(root: Readonly<GModelRoot>): void {
-        this.editorContext.notifyModelRootChanged(root, this);
     }
 }
--- a/packages/client/src/base/editor-context-service.ts
+++ b/packages/client/src/base/editor-context-service.ts
@@ -15,10 +15,8 @@
  ********************************************************************************/
 import {
     Action,
-    AnyObject,
     Args,
     Bounds,
-    CommandStack,
     Disposable,
     DisposableCollection,
     EditMode,
@@ -44,6 +42,7 @@ import {
 import { inject, injectable, postConstruct, preDestroy } from 'inversify';
 import { FocusChange, FocusTracker } from './focus/focus-tracker';
 import { IDiagramOptions, IDiagramStartup } from './model/diagram-loader';
+import { IModelChangeService } from './model/model-change-service';
 import { SelectionChange, SelectionService } from './selection-service';
 
 /**
@@ -85,6 +84,9 @@ export class EditorContextService implements IActionHandler, Disposable, IDiagra
     @inject(SelectionService)
     protected selectionService: SelectionService;
 
+    @inject(TYPES.IModelChangeService)
+    protected modelChangeService: IModelChangeService;
+
     @inject(MousePositionTracker)
     protected mousePositionTracker: MousePositionTracker;
 
@@ -118,13 +120,11 @@ export class EditorContextService implements IActionHandler, Disposable, IDiagra
         return this.onDirtyStateChangedEmitter.event;
     }
 
-    protected _modelRoot?: Readonly<GModelRoot>;
     /**
      * Event that is fired when the model root of the diagram changes i.e. after the `CommandStack` has processed a model update.
      */
-    protected onModelRootChangedEmitter = new Emitter<Readonly<GModelRoot>>();
     get onModelRootChanged(): Event<Readonly<GModelRoot>> {
-        return this.onModelRootChangedEmitter.event;
+        return this.modelChangeService.onModelRootChanged;
     }
 
     /**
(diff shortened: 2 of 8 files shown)


packages/client/src/base/editor-context-service.ts
```diff
@@ -196,6 +205,21 @@ export class EditorContextService implements IActionHandler, Disposable, IDiagra
...
     }
 
+    /**
+     * Notifies the service about a viewport change. This method should not be called
+     * directly. It is called by the `CommandStack` after a viewport modifying command has been processed.
+     * @throws an error if the notifier is not a `CommandStack`
+     * @param viewport the new viewport
+     * @param notifier the object that triggered the viewport change
+     */
+    notifyViewportChanged(viewport: Readonly<Viewport>, notifier: AnyObject): void {
```
martin-fleck-at, line 215: Effectively what we call this with is `GModelRoot & Viewport`, i.e., the result of the `isViewport` typecheck. Since this is only on the client-side, I wonder whether we shouldn't use that type for the listeners instead of just the `Viewport`? Alternatively, we can call this method with just `this.currentViewport` in the command stack.

packages/client/src/base/editor-context-service.ts
```diff
@@ -196,6 +205,21 @@ export class EditorContextService implements IActionHandler, Disposable, IDiagra
...
+        if (!(notifier instanceof CommandStack)) {
+            throw new Error('Invalid viewport change notification. Notifier is not an instance of `CommandStack`.');
```
martin-fleck-at, line 217: Out of curiosity, what is the reason for this restriction? Seems odd that adopters cannot trigger the notification but I see it is the same as in the model root change.

packages/client/src/base/editor-context-service.ts
tortmayr: Fine with me, we can change the type of the event and fire it with this.currentViewport.

packages/glsp-sprotty/src/api-override.ts
```diff
@@ -198,3 +199,53 @@ export interface IActionDispatcher extends SIActionDispatcher {
...
+     */
+    redo(): Promise<GModelRoot>;
+
+    /**
+     * Event that is fired after a command has been successfully executed on the stack. (i.e. after the model has been updated).
+     */
+    onCommandExecuted: Event<CommandExecutionData>;
+}
+
+export type CommandStackProvider = () => Promise<ICommandStack>;
```
martin-fleck-at, line 251: nit: Should we re-add the comments that is in the original code?

Actually, we do not use that interface in our code base at all and also do not offer any default binding for it [like Sprotty does](https://github.com/eclipse-sprotty/sprotty/blob/f9b2433481cc27a1ac0c92d525a92039ae7f6c76/packages/sprotty/src/base/di.config.ts#L81). Should we even export this, should we provide a default binding?

packages/glsp-sprotty/src/api-override.ts
```diff
@@ -198,3 +199,53 @@ export interface IActionDispatcher extends SIActionDispatcher {
...
+export interface ICommandStack extends SICommandStack {
+    /**
+     * Executes the given command on the current model and returns a
+     * Promise for the new result.
+     *
+     * Unless it is a special command, it is pushed to the undo stack
+     * such that it can be rolled back later and the redo stack is
+     * cleared.
+     */
+    execute(command: ICommand): Promise<GModelRoot>;
```
martin-fleck-at, line 222: Any reason you re-specified this method? Seems like we could just take it from the super interface.

packages/glsp-sprotty/src/api-override.ts
```diff
@@ -198,3 +199,53 @@ export interface IActionDispatcher extends SIActionDispatcher {
...
+
+    /**
+     * Executes all of the given commands. As opposed to calling
+     * execute() multiple times, the Viewer is only updated once after
+     * the last command has been executed.
+     */
+    executeAll(commands: ICommand[]): Promise<GModelRoot>;
```
martin-fleck-at, line 229: Any reason you re-specified this method? Seems like we could just take it from the super interface.

packages/glsp-sprotty/src/api-override.ts
```diff
@@ -198,3 +199,53 @@ export interface IActionDispatcher extends SIActionDispatcher {
...
+
+    /**
+     * Client-side undo/redo is not supported in GLSP. The server is responsible for handling undo/redo requests.
+     * This method is required to maintain compatibility with the sprotty API.
+     * Implementation should always be a no-op that returns the current model.
+     */
+    redo(): Promise<GModelRoot>;
+
+    /**
+     * Event that is fired after a command has been successfully executed on the stack. (i.e. after the model has been updated).
```
martin-fleck-at, line 246:
```suggestion
     * Event fired after a command has been successfully executed on the stack (i.e., after the model has been updated).
```

packages/client/src/base/model/model-change-service.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    protected onViewportChangedEmitter = new Emitter<Readonly<Viewport>>();
+    get onViewportChanged(): Event<Readonly<Viewport>> {
+        return this.onViewportChangedEmitter.event;
+    }
+
+    @postConstruct()
+    protected initialize(): void {
+        this.toDispose.push(this.onModelRootChangedEmitter, this.onViewportChangedEmitter);
+        this.commandStack.onCommandExecuted(data => this.handleCommandExecution(data.command, data.newRoot), this);
```
martin-fleck-at, line 87: nit: I think we already have the correct `this` here without giving it as a parameter.

packages/client/src/base/model/model-change-service.ts
```diff
@@ -0,0 +1,144 @@
...
+export class ModelChangeService implements IModelChangeService, Disposable {
+    @inject(LazyInjector)
+    protected lazyInjector: LazyInjector;
+    protected _currentRoot: Readonly<GModelRoot>;
+    protected lastViewport?: Readonly<Viewport>;
+    protected toDispose = new DisposableCollection();
+
+    get currentRoot(): Readonly<GModelRoot> {
+        if (!this._currentRoot) {
+            throw new Error('Model root not available yet');
```
martin-fleck-at, line 65: I know this was the previous behavior, just wondering whether we should not return `| undefined` here and only throw the error in the editor context? Because overall, it would be nice to be able to check somewhere whether root was already set without automatically triggering an error.

packages/client/src/base/model/model-change-service.ts
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Event that is fired when the model root of the diagram changes i.e. after the `CommandStack` has processed a model update.
+     */
+    onModelRootChanged: Event<Readonly<GModelRoot>>;
+
+    /**
+     * Event that is fired when the viewport of the diagram changes i.e. after the `CommandStack` has processed a viewport update.
+     * By default, this event is only fired if the viewport was changed via a `SetViewportCommand` or `BoundsAwareViewportCommand`
+     */
+    onViewportChanged: Event<Readonly<Viewport>>;
```
martin-fleck-at, line 53: I'm wondering whether it wouldn't be more future proof to have an actual interface here instead of just `Readonly<Viewport>`? Like `ViewportChange` because we could also include the old viewport. We already have a bunch of similar events, like `SelectionChange`, `FocusChange`, `ValueChange`, `DirtyStateChange`, etc.

packages/client/src/base/editor-context-service.ts
```diff
@@ -143,6 +143,14 @@ export class EditorContextService implements IActionHandler, Disposable, IDiagra
         return this.selectionService.onSelectionChanged;
     }
 
+    /**
+     * Event that is fired when the viewport of the diagram changes i.e. after the `CommandStack` has processed a viewport update.
+     * By default, this event is only fired if the viewport was changed via a `SetViewportCommand` or `BoundsAwareViewportCommand`
+     */
+    get onViewportChanged(): Event<Readonly<Viewport>> {
```
martin-fleck-at, line 150: I wonder whether we should also introduce a listener type like we have for:
- modelRootChanged -> TYPES.IGModelRootListener
- editModeChanged -> TYPES.IEditModeListener

What do you think?

packages/client/src/base/command-stack.ts
```diff
@@ -15,25 +15,37 @@
...
     ICommand,
-    LazyInjector,
-    SetModelCommand,
-    UpdateModelCommand
+    ICommandStack,
+    LazyInjector
 } from '@eclipse-glsp/sprotty';
-import { inject, injectable, preDestroy } from 'inversify';
+import { inject, injectable, postConstruct, preDestroy } from 'inversify';
+import { Emitter } from 'vscode-jsonrpc';
```
martin-fleck-at, line 29: We definitely want the import from `@eclipse-glsp/sprotty` and not `vscode-jsonrpc`.

martin-fleck-at (changes requested):
Overall very good change, I just have a few comments and questions :-)

packages/glsp-sprotty/src/api-override.ts
tortmayr: I re-specified every method that uses the sprotty SModel type (e.g. SModelRootImpl) instead of our gmodel types (GModelRoot). 

If we override the interface anyway we can also correct the types, witch improves IDE support. (We already do that for other classes/interfaces as well like the mouse listener or the vnode decorator.

packages/client/src/base/model/model-change-service.ts
tortmayr: Fine with me. 
In practice I don't think it really matters.
You will always have a root anyways because it is set in the initialzation phased of the action dispatcher.

packages/client/src/base/command-stack.ts
tortmayr: Fair point

martin-fleck-at (approved):
LGTM!

PR #447, opened 2025-09-17
GLSP-1578: Update publish-next.yml

tortmayr:
#### What it does
Ensure that we are also publish if there are changes in the examples directories

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #448, opened 2025-10-02
Improve wrapping behavior of containers

haydar-metin:
#### What it does

Works on https://github.com/eclipse-glsp/glsp/issues/1384.

This PR is the initial contribution to provide visual feedback that a container/parent wraps a child even while the child is currently moving. In subsequent PR's those containers will be extended and changing the parent itself of children will be implemented.

#### How to test

Move single/multiple selected child elements around or resize them (also with the keyboard). The parent will wrap it.

https://github.com/user-attachments/assets/482a449f-9792-4f86-bf53-6fa3e36f2941

I will revert the changes to the example file before merging.

#### Changelog

- `tracker` variables in the change-bounds files have been renamed to `moveTracker` and `resizeTracker`
- `ChangeBoundsTracker#startTracking` requires the root node as argument

### DIFF
--- a/packages/client/src/features/tools/change-bounds/change-bounds-manager.ts
+++ b/packages/client/src/features/tools/change-bounds/change-bounds-manager.ts
@@ -98,14 +98,14 @@ export interface IChangeBoundsManager {
      * @param size - The size to check.
      * @returns True if the element has a valid size, false otherwise.
      */
-    hasValidSize(element: GModelElement, size?: Dimension): boolean;
+    hasValidSize(element: GModelElement, size?: Dimension, options?: ChangeBoundsManagerSizeOptions): boolean;
 
     /**
      * Get the minimum size of an element for changing bounds.
      * @param element - The element to get the minimum size for.
      * @returns The minimum size of the element.
      */
-    getMinimumSize(element: GModelElement): Dimension;
+    getMinimumSize(element: GModelElement, options?: ChangeBoundsManagerSizeOptions): Dimension;
 
     /**
      * Determine whether to use movement restriction for changing bounds.
@@ -177,6 +177,10 @@ export interface IChangeBoundsManager {
     createTracker(): ChangeBoundsTracker;
 }
 
+export interface ChangeBoundsManagerSizeOptions {
+    useComputedDimensions?: boolean;
+}
+
 /**
  * The default {@link IChangeBoundsManager} implementation. It is responsible for managing
  * the change of bounds for {@link GModelElement}s.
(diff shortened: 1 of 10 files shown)


haydar-metin:
@martin-fleck-at @tortmayr this PR is now ready to review. Just one question, how should we handle invalid arrow key movements? For example, if you use the arrow keys then you can move the elements to overlapping states (on main).

packages/client/src/features/tools/change-bounds/change-bounds-manager.ts
```diff
@@ -98,14 +98,14 @@ export interface IChangeBoundsManager {
...
-    hasValidSize(element: GModelElement, size?: Dimension): boolean;
+    hasValidSize(element: GModelElement, size?: Dimension, options?: ChangeBoundsManagerSizeOptions): boolean;
 
     /**
      * Get the minimum size of an element for changing bounds.
      * @param element - The element to get the minimum size for.
      * @returns The minimum size of the element.
      */
-    getMinimumSize(element: GModelElement): Dimension;
+    getMinimumSize(element: GModelElement, options?: ChangeBoundsManagerSizeOptions): Dimension;
```
martin-fleck-at: nit: missing parameter documentation

packages/client/src/features/tools/change-bounds/change-bounds-tracker.ts
```diff
@@ -442,6 +576,227 @@ export class ChangeBoundsTracker {
...
+
+        // If we have ancestors, we need to calculate the initial bounds based on the ancestors
+        let deltaX = 0;
+        let deltaY = 0;
+        for (const ancestor of ancestors) {
+            const ancestorInitialBounds = initialBounds[ancestor.id] ?? ancestor.bounds;
+            const ancestorBounds = ancestor.bounds;
+
+            deltaX += Bounds.left(ancestorBounds) - Bounds.left(ancestorInitialBounds);
+            deltaY += Bounds.top(ancestorBounds) - Bounds.top(ancestorInitialBounds);
```
martin-fleck-at: very very nit: I do like the utility functions we have to make geometrical calculations a little bit more readable, so we could also write this as
```
const delta = Point.vector(Bounds.position(ancestorInitialBounds), Bounds.position(ancestorBounds));
deltaX += delta.x;
deltaY += delta.y;
```

packages/client/src/features/tools/change-bounds/change-bounds-manager.ts
```diff
@@ -269,6 +273,14 @@ export class ChangeBoundsManager implements IChangeBoundsManager {
         // restriction feedback on each element
         trackedMove.elementMoves.forEach(move => this.addMoveRestrictionFeedback(feedback, move, ctx, event));
 
+        // restriction feedback on each wrapped element (mostly ancestors)
+        Object.values(trackedMove.wrapResizes ?? {}).forEach(elementResize => {
+            this.addMoveRestrictionFeedback(feedback, elementResize, ctx, event);
```
martin-fleck-at: I think it would be great if this handling would be a separate method that can be overwritten.

martin-fleck-at (changes requested):
Hi @haydar-metin! Thank you for that great change! I do have some minor change requests for code style, naming etc. but overall everything looks quite good!

PR #469, opened 2026-02-18
GLSP-1619: Fix AbstractUIExtension error message

### LINKED ISSUE #1619: Error in generating error message "Could not obtain attached parent for initializing UI extension"
https://github.com/eclipse-glsp/glsp-client/blob/master/packages/client/src/base/ui-extension/ui-extension.ts#L76
```throw new Error(`Could not obtain attached parent for initializing UI extension ${this.id}`);```
id is defined in parent class as   `abstract id(): string;`

That combo produces a log entry like this:
`Could not obtain attached parent for initializing UI extension id(){return n.ID}`

I guess there are `()` missing in `Could not obtain attached parent for initializing UI extension ${this.id}`

tortmayr:
#### What it does
Fixes eclipse-glsp/glsp/issues/1619

#### How to test
Trivial fix, no testing necessary.

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #456, opened 2025-12-18
Add support for Model Context Protocol

martin-fleck-at:
#### What it does

MCP client-side support paired with the server-side integration, plus
a generic export pipeline that underpins server-driven export flows
and is reusable by UI flows.

MCP integration
- Adds an opt-in MCP server configuration on initialize and a
  server descriptor on the result.
- Splits the configuration into init- and deploy-controllable parts
  so security-relevant fields stay under deployment control.
- Enable MCP by setting the configuration on initialize; the
  standalone example demonstrates the wiring.

Generic export pipeline
- Format-agnostic export request/response pair, coexisting with the
  legacy SVG-only flow under strict separation.
- Pluggable export-strategy registry with shared post-processing.
- Ships default SVG and PNG strategies; PNG export is a new
  built-in capability.
- Register a custom strategy to add formats or override defaults.

Misc
- Promotes the origin-viewport action to the protocol so the server
  can dispatch it.
- Widens the shutdown contract to allow asynchronous handlers.

Part of https://github.com/eclipse-glsp/glsp/issues/1546

#### How to test

- Nothing breaks
- For the overall change, see [GLSP Node Server](https://github.com/eclipse-glsp/glsp-server-node/pull/120) and [Theia Integration](https://github.com/eclipse-glsp/glsp-theia-integration/pull/271)

#### Changelog

-   [x] This PR should be mentioned in the changelog
-   [ ] This PR introduces a breaking change (if yes, provide more details below for the changelog and the migration guide)

PR #474, opened 2026-03-25
Issues/1546

Sakrafux:
Extends the GLSP MCP Server started in [PR-456](https://github.com/eclipse-glsp/glsp-client/pull/456) according to our meetings @martin-fleck-at.


packages/protocol/src/action-protocol/mcp.ts
```diff
@@ -0,0 +1,125 @@
...
+    export function is(object: unknown): object is ExportPngMcpActionResult {
+        return Action.hasKind(object, KIND) && hasStringProp(object, 'png');
+    }
+
+    export function create(png: string, mcpRequestId: string, options: ExportPngOptions = {}): ExportPngMcpActionResult {
+        return {
+            kind: KIND,
+            mcpRequestId,
+            png,
+            ...options
```
martin-fleck-at: Options are spread here into their nested fields but in the action interface they are declared in a single options field, this will not work as expected.

martin-fleck-at (changes requested):
Thank you @Sakrafux! I think we need to make some changes here but I believe some of the problems will go away once we properly implement https://github.com/eclipse-glsp/glsp/issues/607 which I plan to tackle next, so we can leave that for now as is.

PR #480, opened 2026-04-09
feat: add bidirectional request/response protocol support

martin-fleck-at:
#### What it does

Enable symmetric request/response handling between client and server. Complements the server-side changes in glsp-server-node.

- Add optional `timeout` field to `RequestAction`
- Update `requestUntil()` default to respect `action.timeout`
- Stamp effective timeout onto action in `requestUntil()` so the receiving side can respect it
- Use `request()` vs `requestUntil()` in `handleServerRequest` and `handleClientRequest` based on `action.timeout` presence
- Fix `&&` vs `||` bug in requestId emptiness checks
- Fix `RequestAction` import to resolve GLSP protocol type
- Add async `handleServerRequest()` in `GLSPModelSource` with `RejectAction` error handling
- Add `GetSelectionHandler` for server-initiated selection queries
- Add `GetEditorContextAction` / `SetEditorContextAction` protocol types and handler in `EditorContextService`

Relates to https://github.com/eclipse-glsp/glsp/issues/607

#### How to test

- Try together with https://github.com/eclipse-glsp/glsp-server-node/pull/131
- Add some request actions (selection, editor context, etc.) to a server-side handler and see that you receive the result as expected.

#### Changelog


packages/client/src/features/select/get-selection-handler.ts
```diff
@@ -0,0 +1,30 @@
...
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+import { Action, GetSelectionAction, IActionHandler, SelectionResult } from '@eclipse-glsp/sprotty';
+import { inject, injectable } from 'inversify';
+import { SelectionService } from '../../base/selection-service';
+
+@injectable()
```
tortmayr: Nice catch.
Thanks for adding the missing handler.

packages/protocol/src/action-protocol/contexts.ts
```diff
@@ -85,3 +86,98 @@ export namespace SetContextActions {
...
+}
+
+/**
+ * Response to a {@link GetEditorContextAction} containing a snapshot of the client-side editor state.
+ *
+ * All fields reflect the state at the time of response generation. The server should not assume
+ * that these values are still current when processing the response, as the client state may have
+ * changed in the meantime.
+ */
+export interface SetEditorContextAction extends ResponseAction {
```
tortmayr: The naming of this response action is a bit misleading.
It is not really a set action, its just a (readonly) result response.

The name (to me) suggests that the handling part uses this to update its internal context state. Which is not the intend.

packages/protocol/src/action-protocol/contexts.ts
```diff
@@ -85,3 +86,98 @@ export namespace SetContextActions {
...
+    kind: typeof SetEditorContextAction.KIND;
+
+    /**
+     * The list of currently selected element identifiers.
+     * For a dedicated selection query, use {@link GetSelectionAction} instead.
+     */
+    readonly selectedElementIds: string[];
```
tortmayr: I think it would make sense to extend the editor context object itself for this:

```ts
export interface EditorContext {
    /**
     * The list of selected element identifiers.
     */
    readonly selectedElementIds: string[];

    /**
     * The last recorded mouse position.
     */
    readonly lastMousePosition?: Point;

    /**
     * Custom arguments.
     */
    readonly args?: Args;

     /**
     * The current viewport (scroll position and zoom level).
     * For a dedicated viewport query, use {@link GetViewportAction} instead.
     */
    readonly viewport?: Viewport;

    /**
     * The bounds of the canvas element in the browser.
     * For a dedicated viewport and canvas bounds query, use {@link GetViewportAction} instead.
     */
    readonly canvasBounds?: Bounds;

}
```

The response action itself can than just hold a property with the context object.
The creation methods in the EditorContextService need to be updated as well.

This would make maintaince easier in the long run.
i.e. changes/additions to the EditorContext interace don't need to be synced with the retrieval action.

packages/client/src/features/select/select-module.ts
```diff
@@ -24,6 +33,7 @@ export const selectModule = new FeatureModule(
         configureCommand(context, SelectCommand);
         configureCommand(context, SelectAllCommand);
         configureCommand(context, SelectFeedbackCommand);
+        configureActionHandler(context, GetSelectionAction.KIND, GetSelectionHandler);
```
tortmayr: Side question: Why did you opt for explicit standalone action handler here instead of letting the selection service handle this? Similar to how the EditorContextService handles the get editor context request.

packages/client/src/base/model/glsp-model-source.ts
```diff
@@ -94,6 +98,8 @@ export class GLSPModelSource extends ModelSource implements Disposable {
     @inject(TYPES.IDiagramOptions)
     protected options: IDiagramOptions;
 
+    declare readonly actionDispatcher: GLSPActionDispatcher;
```
tortmayr: Use base interface instead of default implementation here:
```ts
declare readonly actionDispatcher: IActionDispatcher;
```

packages/client/src/base/model/glsp-model-source.ts
```diff
@@ -164,9 +170,32 @@ export class GLSPModelSource extends ModelSource implements Disposable {
...
         this.actionDispatcher.dispatch(action);
     }
 
+    // Fire-and-forget: intentionally not awaited by messageReceived()
+    protected async handleServerRequest(action: RequestAction<ResponseAction>): Promise<void> {
+        try {
+            const response = action.timeout !== undefined
+                ? await this.actionDispatcher.requestUntil(action, action.timeout, true)
+                : await this.actionDispatcher.request(action);
+            if (response) {
```
tortmayr, line 187: Maybe lets move this into a sub method?
So that adopters that need to make changes to the default behavior of forwarding every response to the server have an entrypoint for modifiaction

tortmayr (changes requested):
Thanks Martin! Great new feature.
I did a first high-level review of the code and added some comments.
In addition, I quickly tested the client-server change in combination. From what I can see we don't have any regressions. Everything seems to work as expected.

packages/protocol/src/action-protocol/contexts.ts
martin-fleck-at: You are right, we could go with the `Result` suffix which is already used in other places.

packages/protocol/src/action-protocol/contexts.ts
martin-fleck-at: I was thinking about it and then did not want to make this change larger but it is very non-breaking, so let's just do it! ;-)

packages/client/src/features/select/select-module.ts
martin-fleck-at: Very good question! I think I initially only wanted that handler as a sort of test case so I started out separately. But as you said, we should have handled that anyways. I'll add it to the selection service to be more consistent.

packages/client/src/base/model/glsp-model-source.ts
martin-fleck-at: True, I had the wrong `IActionDispatcher` in mind but we do have an override anyways, good catch, thank you!

packages/client/src/base/model/glsp-model-source.ts
martin-fleck-at: Good idea!

martin-fleck-at:
@tortmayr Thank you! I updated the PR accordingly!

packages/client/src/features/select/select-module.ts
```diff
@@ -33,7 +24,6 @@ export const selectModule = new FeatureModule(
         configureCommand(context, SelectCommand);
         configureCommand(context, SelectAllCommand);
         configureCommand(context, SelectFeedbackCommand);
-        configureActionHandler(context, GetSelectionAction.KIND, GetSelectionHandler);
```
tortmayr: We are missing a registration of the selection service as action handler for the `GetSelectionAction` don't we?

martin-fleck-at: The registration is already in `default.module.ts` (line 127), right next to the other `SelectionService` bindings. I kept it there for consistency with `EditorContextService`, whose `GetEditorContextAction` handler is registered the same way. Happy to move it to `select-module.ts` if you'd prefer the feature-module placement, though.

tortmayr: I see, no thats fine.

tortmayr (approved):
Thanks, all review points are adressed.
The compatibility build with the server PR also succeeded:
https://github.com/eclipse-glsp/glsp-client/actions/runs/24898345067

LGTM! 👍🏼

PR #484, opened 2026-04-10
Update e2e-pr workflow

tortmayr:
- Consolidate PR resolval and remove dependency to third party action
- Add linking step to ensure that the built server uses the correct client protocol package

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #486, opened 2026-04-19
Add support for sticky Manhattan Routing and rounded Edges

### LINKED ISSUE #1652: Sticky Manhattan Router
Based on https://github.com/eclipse-glsp/glsp/discussions/1642, implement a new sticky Manhattan router.

martin-fleck-at:
#### What it does

**Add sticky Manhattan edge router**

Preserves intermediate bend points when a connected node moves and only slides the corner adjacent to the moved endpoint, instead of recomputing the full route like the standard Manhattan router does.

Ported from Ralph Soika's `BPMNManhattanRouter` in Open-BPMN, see https://github.com/eclipse-glsp/glsp/discussions/1642. Opt-in per edge via `edge.routerKind = 'sticky-manhattan'`. Per-edge state lives in a
`WeakMap` keyed on the edge instance. Dedicated anchor computers are registered so the router coexists with the standard Manhattan router.

**Add rounded-corner edge view**

`RoundedCornerEdgeView` extends `GEdgeView` and overrides `createPathForSegments` to emit SVG quadratic curves at every right-angle bend, producing smooth corners on Manhattan-style routes. The radius is clamped per corner to a fraction of the shorter adjacent segment to avoid visual artifacts on short segments.

Ported from Ralph Soika's `BPMNEdgeView`, contributed alongside the sticky Manhattan router via the same discussion. The view works with any router that emits right-angle corners; combining it with `GLSPStickyManhattanEdgeRouter` reproduces the look-and-feel of the original Open-BPMN setup.

sticky-routing

Resolves https://github.com/eclipse-glsp/glsp/issues/1652

#### How to test

- There is a test commit on here that should not be merged but it does adapt the workflow example.

#### Changelog


martin-fleck-at:
Test failures probably related to the change of the workflow example. Once the code is reviewed, I'll remove that commit and make sure the CI passes.

packages/client/src/features/routing/sticky-manhattan-edge-router.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+    container.load(routingModule);
+    const registry = container.get<EdgeRouterRegistry>(EdgeRouterRegistry);
+    const router = registry.get(GLSPStickyManhattanEdgeRouter.KIND) as GLSPStickyManhattanEdgeRouter;
+    return { graph, edge, router };
+}
+
+describe('GLSPStickyManhattanEdgeRouter', () => {
+    describe('KIND', () => {
+        it('uses the sticky-manhattan identifier to avoid collision with the standard Manhattan router', () => {
+            expect(GLSPStickyManhattanEdgeRouter.KIND).to.equal('sticky-manhattan');
```
tortmayr: Unnecessary test of a constant.
Please remove

packages/client/src/features/routing/sticky-manhattan-edge-router.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+        });
+    });
+
+    describe('anchor registration', () => {
```
tortmayr: I'm not sure if we should test this. This is basically just verifying correct inversify hookup.
If we want to test it should probably be in a separate test. It rather tests wiring of the routingmodule than the sticky router itself.
I personally would opt for removing this test.

packages/client/src/features/routing/sticky-manhattan-edge-router.ts
```diff
@@ -0,0 +1,667 @@
...
+ * adjacent to the moved endpoint. Intermediate bend points are kept in place, which
+ * results in more predictable routing when users reposition nodes in complex
+ * diagrams.
+ *
+ * Opt-in per edge by setting `edge.routerKind = 'sticky-manhattan'`.
+ *
+ * Based on the original `BPMNManhattanRouter` by Ralph Soika (Imixs Software
+ * Solutions GmbH), contributed via https://github.com/eclipse-glsp/glsp/discussions/1642.
+ */
+@injectable()
```
tortmayr, line 58: Maybe we should explicitly mention that the router is currently experimental i.e.
API surface and implementation details might change with future improvements (also potentially breaking)

packages/client/src/features/routing/sticky-manhattan-edge-router.ts
```diff
@@ -0,0 +1,667 @@
...
+            edge.targetAnchorCorrection
+        );
+        if (!sourceAnchor || !targetAnchor) {
+            return;
+        }
+
+        if (points.length === 1) {
+            // Classify the first segment via the source side closest to the bend — this
+            // matches how the Manhattan anchor computer picks the source anchor.
+            // The original `BPMNManhattanRouter` used a tolerance-based axis check
```
tortmayr: IMO this comment adds lots of unnecessary noise. Just keep the first sentence, remove the rest.

packages/client/src/views/rounded-corner-edge-view.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+class TestableRoundedCornerEdgeView extends RoundedCornerEdgeView {
+    public override createPathForSegments(segments: Point[]): string {
+        return super.createPathForSegments(segments);
+    }
+    public override computeMaxRadius(corner: Point, prev: Point, next: Point): number {
+        return super.computeMaxRadius(corner, prev, next);
+    }
+}
+
+describe('RoundedCornerEdgeView', () => {
```
tortmayr: Should we really test views on a unit level? We don't do this for any other views.
To me the implementation looks rather brittle heavily relies on implementation/render detatils.
I.e tests might break on minor rendering changes.

If we want to test rendering/views it might be better to do it on an e2e level.

packages/protocol/src/sprotty-geometry-point.ts
```diff
@@ -113,6 +113,28 @@ declare module 'sprotty-protocol/lib/utils/geometry' {
...
+        /**
+         * Returns whether the segment between two points is vertical, i.e. the
+         * points share (approximately) the same x coordinate.
+         *
+         * @param a first point
+         * @param b second point
+         * @param eps tolerance on the x axis (defaults to 1 to match the sub-pixel drift
+         * typical for Manhattan-routed coordinates)
+         */
+        function isVertical(a: Point, b: Point, eps?: number): boolean;
```
tortmayr: We now have a tolerance difference for the default value of this method in contrast to other sprotty builtin like `Point.isAlmostEqual`. (1e-3)

Imo, the tolerance drift of 1px is an implementation detail of the sticky router.
-> Use same default as Point.isAlmostEqual and explicitly specify the 1px eps in the stick router usages.

tortmayr (changes requested):
@martin-fleck-at Thanks for preparing this contribution.
In manual testing everything looks fine and seems to work as expected.
Overall the new router provides a way better UX (at least for my taste) than the default manhattan router.

We should mark the router as experimental. And maybe add a follow-up to reiterate over the initial implementation to see if there is potential for improvements/cleanups.

packages/client/src/features/routing/sticky-manhattan-edge-router.spec.ts
martin-fleck-at: Right!

packages/client/src/features/routing/sticky-manhattan-edge-router.spec.ts
martin-fleck-at: Probably not worth it as other things will fail, I'll remove the test.

packages/client/src/features/routing/sticky-manhattan-edge-router.ts
martin-fleck-at: Good idea, I marked it experimental!

packages/client/src/features/routing/sticky-manhattan-edge-router.ts
martin-fleck-at: Done.

packages/client/src/views/rounded-corner-edge-view.spec.ts
martin-fleck-at: There were some test functions that may have been a little bit interesting but also not too much. I guess it is better to stay consistent, I removed the test again.

packages/protocol/src/sprotty-geometry-point.ts
martin-fleck-at: Done, I introduced a constant for the almost equal value since sprotty does not have that and use that as default and let the user configure the tolerance through the options but defaulting to 1 in the sticky router.

martin-fleck-at:
> @martin-fleck-at Thanks for preparing this contribution. In manual testing everything looks fine and seems to work as expected. Overall the new router provides a way better UX (at least for my taste) than the default manhattan router.
> 
> We should mark the router as experimental. And maybe add a follow-up to reiterate over the initial implementation to see if there is potential for improvements/cleanups.

Thank you! I also like the new router a lot and I already re-worked the code a bit for some clean up. Overall let's see what adopters report.

martin-fleck-at:
@tortmayr Test failure is just for the workflow example override and will be fixed once we remove that agian.

tortmayr (approved):
LGTM! 👍🏼

tortmayr:
Please remove the temp changes to the workflow example so that the ci passes again.

martin-fleck-at:
Done, I removed the workflow example adaptations and the CI runs through now.

PR #495, opened 2026-04-30
GLSP-1657: Remove dependency to lodash

### LINKED ISSUE #1657: Remove direct dependency to lodash
## Summary

The `@eclipse-glsp/client` package has a direct dependency on `lodash` (`4.17.23`) and `@types/lodash`. However, we only use a very minimal subset of lodash utilities (`debounce`, `groupBy`, `partition`, `isEqual`, and `DebouncedFunc`), which does not warrant pulling in the entire library as a dependency.

## Proposed Changes

- Remove `lodash` from `@eclipse-glsp/client` dependencies and `@types/lodash` from both root and client devDependencies
- Add lightweight, self-contained utility replacements in `@eclipse-glsp/protocol`:
  -
## Motivation

- Reduces bundle size by eliminating a large transitive dependency
- Reduces supply chain surface area
- The handful of lodash functions we actually use are straightforward to implement and maintain ourselves

tortmayr:
#### What it does
- Remove direct dependency to lodash
- Introduce replacement utility functions for the small subset that we actually used
- Implement tests for new and existing utility functions

Also:
- Update publish.yml to skip next publishes for dependabot commits

Fixes eclipse-glsp/glsp#1657

#### Changelog

### DIFF
--- a/packages/client/src/features/helper-lines/helper-line-feedback.ts
+++ b/packages/client/src/features/helper-lines/helper-line-feedback.ts
@@ -29,9 +29,9 @@ import {
     Viewport,
     equalUpTo,
     isBoundsAware,
-    partition,
     isDecoration,
-    isViewport
+    isViewport,
+    partition
 } from '@eclipse-glsp/sprotty';
 import { inject, injectable } from 'inversify';
 import '../../../css/helper-lines.css';
--- a/packages/client/src/features/helper-lines/helper-line-feedback.ts
+++ b/packages/client/src/features/helper-lines/helper-line-feedback.ts
@@ -29,11 +29,11 @@ import {
     Viewport,
     equalUpTo,
     isBoundsAware,
+    partition,
     isDecoration,
     isViewport
 } from '@eclipse-glsp/sprotty';
 import { inject, injectable } from 'inversify';
-import { partition } from 'lodash';
 import '../../../css/helper-lines.css';
 import { EditorContextService } from '../../base/editor-context-service';
 import { FeedbackCommand } from '../../base/feedback/feedback-command';
@@ -125,7 +125,9 @@ export class DrawHelperLinesFeedbackCommand extends FeedbackCommand {
         removeSelectionBounds(context.root);
         const alignableElements = getMatchingElements(context.root.index, this.isAlignableElementPredicate);
         this.log('All alignable elements: ', alignableElements);
-        const [referenceElements, elements] = partition(alignableElements, element => this.elementIds.includes(element.id));
+        const { match: referenceElements, rest: elements } = partition(alignableElements, (element): element is BoundsAwareModelElement =>
+            this.elementIds.includes(element.id)
+        );
         this.log('Split alignable elements into reference elements and other elements: ', referenceElements, elements);
         if (referenceElements.length === 0) {
             this.log('--> No helper lines as we do not have any reference elements.');
--- a/packages/client/src/features/search-palette/search-palette.ts
+++ b/packages/client/src/features/search-palette/search-palette.ts
@@ -14,9 +14,8 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import { GModelElement, GModelRoot, GNode, LabeledAction, SelectAllAction, toArray, TYPES } from '@eclipse-glsp/sprotty';
+import { GModelElement, GModelRoot, GNode, LabeledAction, SelectAllAction, debounce, toArray, TYPES } from '@eclipse-glsp/sprotty';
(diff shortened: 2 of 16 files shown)


packages/client/src/features/search-palette/search-palette.ts
```diff
@@ -150,7 +149,7 @@ export class SearchAutocompletePalette extends BaseAutocompletePalette {
     }
 
     protected getSuggestionsFromLabeledActions(labeledActions: LabeledAction[]): SearchAutocompleteSuggestion[] {
-        return this.cachedSuggestions.filter(c => labeledActions.find(s => isEqual(s, c.action)));
+        return this.cachedSuggestions.filter(c => labeledActions.find(s => s === c.action));
```
martin-fleck-at, line 152: This is a behavior change worth verifying: `isEqual(s, c.action)` was a deep equality check, and `s === c.action` is reference equality. They are only equivalent if `labeledActions` and `this.cachedSuggestions[i].action` always point to the same object instance.

packages/protocol/src/utils/function-util.ts
```diff
@@ -0,0 +1,72 @@
...
+    (...args: Parameters<T>): void;
+    cancel(): void;
+}
+
+export interface DebounceSettings {
+    leading?: boolean;
+    trailing?: boolean;
+}
+
+export function debounce<T extends (...args: any[]) => any>(func: T, wait: number, options: DebounceSettings = {}): DebouncedFunc<T> {
```
martin-fleck-at, line 36: Two small compatibility deltas vs `lodash.DebouncedFunc<T>` worth flagging, since `DebouncedFunc` is re-exported as part of the public protocol surface:
                                                                                                                                                                                                                                                       
  1. The debounced call returns `void` here; lodash returns `ReturnType<T> | undefined`. Anyone reading the return value of a debounced call will break silently on upgrade.                                                                            
  2. No `flush()`, no `pending()`, no `maxWait`. `flush()` in particular is commonly used and is only a few lines to add.                                                                                                                               
                                                                                                                                                                                                                                                        
Suggestion: add a short JSDoc on `debounce` (and/or the file header) explicitly stating this is a minimal lodash-compatible subset, listing what is intentionally not supported. That gives downstream adopters a clear contract instead of having them discover the gaps at runtime.

packages/client/src/features/helper-lines/helper-line-feedback.ts
```diff
@@ -29,11 +29,11 @@ import {
     Viewport,
     equalUpTo,
     isBoundsAware,
+    partition,
```
martin-fleck-at: Nit: `partition` is inserted between `isBoundsAware` and `isDecoration`, which breaks the alphabetical import ordering used in the rest of this block.

martin-fleck-at (changes requested):
Thanks for the cleanup — the additional tests for `groupBy`, `partition`, and `debounce` are a nice byproduct. A few comments inline. The most important is on `search-palette.ts` — swapping `isEqual` for `===` looks like a semantic regression worth verifying before merge.

One general note: since `DebouncedFunc`, `debounce`, `groupBy`, and `partition` all become part of the public `@eclipse-glsp/protocol` API on merge, it would be worth a short paragraph in the changelog calling out that they are a minimal lodash-compatible subset (with the deltas mentioned in the comments). Adopters migrating off lodash types in their own code will want to know what is missing before they switch.

packages/client/src/features/search-palette/search-palette.ts
tortmayr: Afaik they do. so we don't need deepEquality here.
But I can add it it you wan.t

packages/protocol/src/utils/function-util.ts
tortmayr: Sounds good

packages/client/src/features/search-palette/search-palette.ts
tortmayr: Double checkd:
Reference equality is fine here. At line 74, retrieveSuggestions returns suggestions.map(s => s.action) — those are the same LabeledAction object references stored in cachedSuggestions[i].action. The autocomplete widget then passes back a subset of
  those same objects to visibleSuggestionsChanged / getSuggestionsFromLabeledActions, so it's always the same instances being compared.

  No need to add an ID — the === check is correct

packages/client/src/features/helper-lines/helper-line-feedback.ts
tortmayr: done

packages/protocol/src/utils/function-util.ts
tortmayr: done

martin-fleck-at (approved):
LGTM!

PR #500, opened 2026-05-02
GLSP-1656: Rework server bundle consumption and unify start scripts

tortmayr:
#### What it does
Remove the direct dependency on
@eclipse-glsp-examples/workflow-server-bundled-web to break the circular dependency between glsp-client and glsp-server-node.

- Download server bundles via npm pack at script time instead of as a dependency
- Add shared download-server.ts utility with version caching
- Replace dev.ts and start-example-server.ts with unified start.ts
- Support --browser, --dev, --client-only, --no-open, --port, --host flags
- Remove bundling from prepare script
- Update webpack config to resolve web bundle from server/ directory
- Add explicit bundle step to E2E and deployment workflows
- Update READMEs
- 

#### How to test

-   [x]  building with yarn works
-   [x]  yarn start  works (start nodeServer, webserver and open url)
-   [x]  yarn start:browser works (start  webserver and open url)
-   [x]  yarn start --client-only fails (start webserver, cannot load because server has not been started)
 -  [x]  yarn dev  works (picks up changes in source code)
 -  [x] yarn dev:browser works
 
#### Changelog


martin-fleck-at (approved):
LGTM!

PR #504, opened 2026-05-20
GLSP-31: Update workflow example (standalone)

### LINKED ISSUE #31: Update workflow standalone example

tortmayr:
#### What it does

GLSP-31: Update workflow example (standalone)

##### Migrate bundling to esbuild
Part-of: eclipse-glsp/glsp#1693

##### Rework workflow-standalone example
- Redesign standalone example app
  - Central diagram widget
  - Proper title bar
  - Additional toggleable section for keyboard bindings
Add/introduce following features
- Context menu support
  - Server-side provided context menus
  - Navigation/Marker menu entries
  - Edit task menu entries
- Title bar with
  - Diagram menu similar to Theia integration
    - Including layout commands
  - Dirty state marker
  - Undo/Redo buttons
- Theming support
  - 5 seletable themes
  - light/dark mode variants
  - Theme switcher in title bar
- Extended urlParameter configuration
  - grid,readonly,mcp,theme,mode

Fixes eclipse-glsp/glsp#31
##### Additional fixes
- Fixes an issue with the GLSP Projection view which did not calculate
scrollbars correctly for the diagrams that are off-center (don't start
at 0,0)

- Fixes an issue with the GLSPWebSocketConnectionHandler which did not
properly try to reconnect if it was launched before the server was
available

- Add comment section to claude.md
#### Changelog

    -Workflow-example redesign is not part of the changelog, but the fix for the projection view should be part of it
   - WebSocketConnectionProvider fix should be mentioned. Might be affecting some edges cases of adopters that relied on the reconnection behavior not triggering if the server is not available yet


tortmayr:
@martin-fleck-at Could you please review this change. Sorry, its quite large 😉 

@planger @haydar-metin If you have time please test the reworked standalone example.
Any Feedback is appreciated [Open browser example](https://eclipse-glsp.github.io/glsp-client/pr-previews/pr-504//diagram.html)

planger (approved):
Wow this is just wonderful! Very very beautiful, modern, but not overdone.
Excellent work, thank you very much!

haydar-metin:
image

That is a lie!

---

image

The shadow at the bottom is cut off.

---

Just an opinion without a solution:
image

the scrollbars are sometimes hard to see.

---

I'm not sure if it is fixable:

You can not move the scrollbar if the cursor is outside of the frame.

---

image

Not sure if we should fix that :D 

---

Amazing work! Looks soooo much better now.

examples/workflow-standalone/app/diagram.html
```diff
@@ -1,32 +1,488 @@
...
+                <div class="card-canvas">
+                    <div id="sprotty"></div>
+                </div>
+
+                <footer class="card-footer">
+                    <div class="shortcuts-toggle" onclick="toggleShortcuts()">
+                        <span class="shortcuts-chevron" id="shortcutsChevron">&#x25B8;</span>
+                        Keyboard Shortcuts
+                    </div>
+                    <div class="shortcuts-panel" id="shortcutsPanel">
```
haydar-metin, line 259: We should implement a registry for all possible shortcuts as a follow up (I'm not sure if we have already something similar with the accessibility implementation). This would also allow us to identify if two shortcuts are bound in the same context.

examples/workflow-standalone/src/common/features/context-menu/workflow-standalone-context-menu-provider.ts
```diff
@@ -0,0 +1,106 @@
...
+                      ]
+                    : [],
+            isEnabled: () => !this.editorContext.isReadonly && selectedTasks.length === 1
+        };
+
+        return Promise.resolve([editTask, goTo]);
+    }
+}
+
+export function collectIssueMarkers(root: GParentElement): GIssueMarker[] {
```
haydar-metin, line 97: Sounds like something that should be moved to the framework instead of examples? What do you think?

examples/workflow-standalone/app/diagram.html
tortmayr: Yeah, we should do this as a follow-up issue

examples/workflow-standalone/src/common/features/context-menu/workflow-standalone-context-menu-provider.ts
tortmayr: Not sure. I currently don't see any reuse for this. 
Also its not a very complex function after all, not really a point to share it globally proactivly

tortmayr:
Thanks for the review @haydar-metin. 

I fixed the first three mentioned issue. The other two are not easily fixable, and i think the behaviors is okay for an example app.

haydar-metin (approved):

PR #505, opened 2026-05-21
GLSP-1675: Avoid unnecessary await for sync action handlers

### LINKED ISSUE #1675: [Regression] Helper line snapping seems to be broken
https://github.com/user-attachments/assets/7d7306cf-33be-4b03-8bfb-3b04a651dd5d

In the workflow example the helper line snapping no longer works correctly.

Especially for vertical snapping it often snaps with a "offset" instead of correctly aligning with the elements we snapped to.

Info
- 2.6.0 Release: also broken
- 2.5.0 Relase: still working as expected

tortmayr:
Fixes https://github.com/eclipse-glsp/glsp/issues/1675

#### What it does

- Only await handler results that are actually promises
- Sync handler results are used directly, preserving sprotty's original microtask timing for the common case
#### How to test

- Test snapping in the deployed example and ensure that it now snaps again as expected.
  Proper execution of sync handlers without microtasking roundtrip should resolve the timing issues)
#### Changelog

### DIFF
--- a/packages/client/src/base/action-dispatcher.ts
+++ b/packages/client/src/base/action-dispatcher.ts
@@ -22,6 +22,7 @@ import {
     EMPTY_ROOT,
     GModelRoot,
     IActionDispatcher,
+    MaybePromise,
     RejectAction,
     RequestAction,
     ResponseAction,
@@ -184,7 +185,7 @@ export class GLSPActionDispatcher extends ActionDispatcher implements IGModelRoo
         const handlerResults: Promise<any>[] = [];
         for (const handler of handlers) {
             const maybeResult = handler.handle(action);
-            const result = this.isPromiseLike(maybeResult) ? await maybeResult : maybeResult;
+            const result = MaybePromise.isPromise(maybeResult) ? await maybeResult : maybeResult;
             if (Action.is(result)) {
                 handlerResults.push(this.dispatch(result));
             } else if (result !== undefined) {
@@ -193,10 +194,6 @@ export class GLSPActionDispatcher extends ActionDispatcher implements IGModelRoo
             }
         }
         await Promise.all(handlerResults);
-    }
-
-    protected isPromiseLike(value: unknown): value is PromiseLike<unknown> {
-        return typeof value === 'object' && value !== undefined && typeof (value as any).then === 'function';
     }
 
     override request<Res extends ResponseAction>(action: RequestAction<Res>): Promise<Res> {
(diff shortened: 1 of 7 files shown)


packages/client/src/base/action-dispatcher.ts
```diff
@@ -194,6 +195,10 @@ export class GLSPActionDispatcher extends ActionDispatcher implements IGModelRoo
         await Promise.all(handlerResults);
     }
 
+    protected isPromiseLike(value: unknown): value is PromiseLike<unknown> {
+        return typeof value === 'object' && value !== undefined && typeof (value as any).then === 'function';
+    }
```
martin-fleck-at: Here we would run into the infamous JS quirk is that `typeof null === 'object'` is true and it would crash when we read `then`. So we should probably go with something like:

```
protected isPromiseLike(value: unknown): value is PromiseLike<unknown> {
      return value != null && typeof value === 'object' && typeof (value as { then?: unknown }).then === 'function';
  }
```

This seems like such a global function, should we rather add it to a utility somewhere? Do we also need it on the server side then?

martin-fleck-at (changes requested):

packages/client/src/base/action-dispatcher.ts
tortmayr: Good catch. I moved the entire function into our type util as typeguard for maybepromises (`MaybePromise.isPromise()`)

martin-fleck-at: Yes, much much better! The only nitpick I have is that the parameter is now `value: MaybePromise<T>` while most of our other type guards take `object: unknown`. I see the motivation but also there is no reason to restrict it, I believe.

tortmayr: There definitely is.
- With an unkown value we lose typesaftey. i.e we don't know the T type of the maybe promise
- With an unknown value we introduce potential for false positives. i.e. any object that has a then function property passes the check

Also its not a classical `is/type of` typeguard i.e. its not supposed to check wether and unknown value  is a `MaybePromise`.

Its dedicated helper function to detect whether a given MaybePromise is a Promise or sync. 

object:unkown does not make any sense here IMO

martin-fleck-at: Sure, I'm happy either way then.

martin-fleck-at (approved):
LGTM!

PR #506, opened 2026-05-27
GLSP-1656: Update start scripts

tortmayr:
#### What it does
Update start scritps to allso allow starting the example server with an external bundle e.g. a local node-server change
Allow configuration of webserver port

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #507, opened 2026-05-27
GLSP-1609: Restore parallel-start dispatch for async action handlers

martin-fleck-at:
#### What it does

- Invoke all handlers in one synchronous burst (sprotty parallel-start)
- Await handlers' follow-up work collectively, not each one in the loop
- Stop a slow or stuck async handler from starving siblings or deadlocking
- Extract processHandlerResult; keep blockUntil set during the burst
- Reject dispatch promise for handler-less requests (sprotty parity)
- Clear requestUntil timeouts on all settle paths to fix a timer leak
- Replace the sequential test; add starvation, deadlock and leak tests

Relates to https://github.com/eclipse-glsp/glsp/issues/1609

#### How to test

- Check if all tests run or whether there is a scenario that is missing.

#### Changelog

### DIFF
--- a/packages/client/src/base/action-dispatcher.ts
+++ b/packages/client/src/base/action-dispatcher.ts
@@ -21,6 +21,7 @@ import {
     Deferred,
     EMPTY_ROOT,
     GModelRoot,
+    HandleActionResult,
     IActionDispatcher,
     MaybePromise,
     RejectAction,
@@ -133,17 +134,15 @@ export class GLSPActionDispatcher extends ActionDispatcher implements IGModelRoo
     }
 
     protected async handleResponseAction(action: ResponseAction): Promise<void> {
-        // clear any pending timeout
         const timeout = this.timeouts.get(action.responseId);
         if (timeout !== undefined) {
             clearTimeout(timeout);
             this.timeouts.delete(action.responseId);
         }
 
-        // check for matching request
         const request = this.requests.get(action.responseId);
         if (!request) {
-            // treat no matching request by dispatching action normally
+            // No pending request: re-dispatch as a normal action.
             this.logger.log(this, 'No matching request for response, dispatch normally', action);
             action.responseId = '';
             return this.handleAction(action);
@@ -151,7 +150,6 @@ export class GLSPActionDispatcher extends ActionDispatcher implements IGModelRoo
 
         this.requests.delete(action.responseId);
         if (RejectAction.is(action)) {
-            // translation reject action to request rejection
             request.reject(new Error(action.message));
             this.logger.warn(this, `Request with id ${action.responseId} failed.`, action.message, action.detail);
         } else {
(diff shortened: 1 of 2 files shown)


packages/client/src/base/action-dispatcher.ts
```diff
@@ -170,35 +168,56 @@ export class GLSPActionDispatcher extends ActionDispatcher implements IGModelRoo
...
         }
+        // Reject the dispatch promise too (sprotty parity), so direct dispatch() callers also see the
+        // failure, not only request() callers via the deferred.
+        throw error;
     }
 
     protected async handleActionWithHandler(action: Action, handlers: ActionHandler[]): Promise<void> {
         this.logger.log(this, 'Handle', action);
-        const handlerResults: Promise<any>[] = [];
+        // No `await` inside the loop: all handlers are invoked in one uninterrupted burst that cannot
```
tortmayr: I guess there is no need for a multiline documentation block here if we restore the original behavior 😉

packages/client/src/base/action-dispatcher.ts
```diff
@@ -170,35 +168,56 @@ export class GLSPActionDispatcher extends ActionDispatcher implements IGModelRoo
...
-            this.requests.delete(action.requestId);
-            request.reject(error);
+        if (RequestAction.is(action)) {
+            const request = this.requests.get(action.requestId);
+            if (request !== undefined) {
+                this.requests.delete(action.requestId);
+                request.reject(error);
+            }
         }
+        // Reject the dispatch promise too (sprotty parity), so direct dispatch() callers also see the
```
tortmayr: Really necessary to comment that? rejection via throw is pretty self explanatory

packages/client/src/base/action-dispatcher.ts
```diff
@@ -170,35 +168,56 @@ export class GLSPActionDispatcher extends ActionDispatcher implements IGModelRoo
...
+                if (resultPromise) {
+                    handlerResults.push(resultPromise);
+                }
             }
         }
         await Promise.all(handlerResults);
     }
 
+    /**
+     * `blockUntil` is set here; for async handlers that happens only after they resolve, not during
```
tortmayr: Should this be an actual inline comment .
Its defacto the same behavior as before and as it in the sprotty base. Not sure if we really need to document this.

tortmayr (approved):
Looks good to me. 
Inline comments are mostly doc nitpicks

packages/client/src/base/action-dispatcher.ts
martin-fleck-at: Good catch, I will fix it.

packages/client/src/base/action-dispatcher.ts
martin-fleck-at: Good catch, I will fix it.

packages/client/src/base/action-dispatcher.ts
martin-fleck-at: Good catch, I will fix it.

PR #508, opened 2026-05-27
Make MCP server startup configurable via URL parameters

tortmayr:
#### What it does
- Disable MCP initialization by default in node mode
- Enable MCP with `?mcp` URL parameter
- Replace hand-rolled URL parser with `URLSearchParams`
- Add `--mcp` flag to start.ts, forwarded as webpack env
- Add `start:mcp` and `dev:mcp` scripts

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #512, opened 2026-05-29
Fix e2e workflows

tortmayr:
#### What it does
Fix e2e workflows
and add path ignores for metadata only changes

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #519, opened 2026-06-16
GLSP-1700: Introduce uuid helper

### LINKED ISSUE #1700: Centralize uuid generation
Currently we have no central helper/utility class to create uuids.
This means every GLSP package that needs a uuid has to declare its own dependency to a uuid provider package. 
This typically causes drift and we end up with different used uuid dependency versions across glsp packages.
This makes it hard to maintain and audit for security fixes.

Proposal:

Introduce a uuid utility module in `@eclipse-glsp/protocol`.  This  can then be consistently reused 
by all glsp packages that need uuids.
As a result we have on central entry point for declaring the uuid dependency and can maintain/audit it easily.
Consuming packages no longer need to worry about uuid dependencies at all.
In addition we can enforce proper usage of the helper by restricting uuid imports via eslint.

tortmayr:
#### What it does
- Add shared uuid util to protocol package. Consuming packages should use the helper when needing uuids instead of depending on their own uuid version
- Update code-base to consistently use new helper
- Update eslint config to restrict direct 'uuid' imports and adjust import rules for examples

Fixes eclipse-glsp/glsp#1700

#### Changelog

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -47,6 +47,14 @@ export default [
                     name: 'sprotty-protocol',
                     message:
                         "The sprotty-protocol default exports are customized and reexported by GLSP. Please use '@eclipse-glsp/client' instead"
+                },
+                {
+                    name: 'uuid',
+                    message: "Use the 'generateUuid'/'isUuid' helpers (from '@eclipse-glsp/protocol') instead of importing 'uuid' directly."
+                },
+                {
+                    name: 'uuid/*',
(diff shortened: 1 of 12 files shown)


eslint.config.mjs
```diff
@@ -75,14 +83,14 @@ export default [
                     message: "The protocol package should not have any direct 'sprotty' dependencies. Try to use 'sprotty-protocol' instead"
                 },
                 {
-                    name: '@eclipse-glsp/client',
```
haydar-metin, line 78: I assume the protocol package doesnt have a reference to `@eclipse-glsp/client` as dependency and therefore we can remove it?

haydar-metin (approved):
LGTM

PR #525, opened 2026-07-15
GLSP-1717: Fix bounds computation error after diagram export

### LINKED ISSUE #1717: Compute bounds after export throws exception for deleted element
1. Trigger export
2. Remove an element from the model (or have changing ids)
3. Modul update fails with
```
Uncaught exception for action 'computedBounds'
org.eclipse.glsp.server.types.GLSPServerException: Model element not found! ID: ....
```

The export triggers the HiddenBoundsUpdater, which collects the bounds in a map but not clearing them. When computing the bounds again, the deleted element bounds are still present (other ones are overridden). The ComputedBoundsActionHandler will throw the exception above.

tortmayr:
#### What it does

Exports render the diagram through the hidden viewer, so the `GLSPHiddenBoundsUpdater`
collected bounds and routing points for the export clone but never cleaned them up.
The stale entries leaked into the next `computedBounds` action, causing
`Model element not found` server errors when elements had been deleted in the meantime.

- Always reset the collected bounds data and routing points at the end of `postUpdate`
  (via try/finally), regardless of the rendering cause or errors during the computation

Fixes eclipse-glsp/glsp#1717

#### How to test

```bash
pnpm install && pnpm build
pnpm -C examples/workflow-standalone start
```

1. Select a task node **first** (selecting an element after the export triggers a local
   bounds computation that resets the leaked state and masks the bug)
2. Trigger the SVG export with `Ctrl+Shift+E`
3. Delete the selected node with `Del`

Without this fix, the `computedBounds` response contains stale entries for every element
of the export clone (including the deleted one) and the server rejects the update with
`Could not retrieve element with id: 'task_ChkTp_icon'`. With the fix, the update succeeds.

- [x] Verified locally against the workflow example (node server)

#### Follow-ups

The same early-return-without-cleanup flaw exists upstream in sprotty's
`HiddenBoundsUpdater`, so plain-sprotty adopters are affected too. Worth filing upstream.

#### Changelog

### DIFF
--- a/packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts
+++ b/packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts
@@ -67,74 +67,81 @@ export class GLSPHiddenBoundsUpdater extends HiddenBoundsUpdater {
     }
 
     override postUpdate(cause?: Action): void {
-        if (cause === undefined || cause.kind !== RequestBoundsAction.KIND) {
-            return;
-        }
+        try {
+            if (cause === undefined || cause.kind !== RequestBoundsAction.KIND) {
+                return;
+            }
 
-        if (LocalRequestBoundsAction.is(cause) && cause.elementIDs) {
-            this.focusOnElements(cause.elementIDs);
-        }
+            if (LocalRequestBoundsAction.is(cause) && cause.elementIDs) {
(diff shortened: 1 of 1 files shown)


haydar-metin:
Thanks github-actions for the link :)) 

Tested + LGTM!

PR #531, opened 2026-08-19
Issues/1725 client side feedback element bounds

### LINKED ISSUE #1725: Bounds of client-side feedback elements are reported to the server
Validation markers and the dangling edge of the edge-creation tool only exist on the client, but they are part of the hidden rendering and therefore end up in the `computedBounds` action. The server cannot resolve their ids and rejects the whole action, so the diagram stops settling.

Reported in #1724 with the node server and the VS Code integration:

```
Could not retrieve element with id: '<random 8 character id>'
```

**How it surfaces**

A `GIssueMarker` is added as client-side feedback by the `ApplyMarkersCommand` and never gets an id assigned, so the client model index generates a random one for it. It is a `GDecoration` and therefore has the `boundsFeature`, so the bounds pass measures it and reports its id like that of any other element.

This normally stays invisible. The hidden model of a server-triggered `RequestBoundsAction` is created from the schema the server just sent and contains no markers, and the result of a client-triggered `LocalRequestBoundsAction` is never forwarded. A marker id only reaches the server when the bounds collected for one hidden rendering leak into the next one, which is what #1717 was about. Two of those paths are left.

- Before glsp-client#525, an export or a bounds pass that fails in between left the collected data behind. That fix is on master but not released yet, so adopters on 2.7.0 still have it.
- Even with the fix, the cleanup lives in `postUpdate`, which the viewer never reaches when a view or the vdom patch throws during the rendering.

The dangling feedback edge of the edge-creation tool has the same exposure. It is created from the adopter's own edge type, so nothing identifies it as feedback, and its route is collected like the route of any real edge.

**Why the client should not report them in the first place**

An id the server does not know is a protocol violation, and the server rejects the entire `computedBounds` for it, which is fair enough. That makes the client the place to fix this. Both remaining leak paths are timing accidents, so treating the cleanup as the only defense keeps the failure one refactor away, whereas never naming a client-only element rules it out by construction. `feedbackFeature` already exists for exactly that concept, but it is currently only ever set and never read.

**What to change**

- [ ] Skip feedback elements and their routes when answering a server bounds request
- [ ] Mark `GIssueMarker` and the dangling feedback edge as feedback elements
- [ ] Reset the collected bounds when a rendering starts rather than only when `postUpdate` completes
- [ ] Add the missing changelog entry for glsp-client#525

martin-fleck-at:
#### What it does

Validation markers and the dangling edge of the edge-creation tool only exist on the client,
but they are part of the hidden rendering and therefore of the `computedBounds` action. Their
ids are unknown to the server, which rejects the whole action, so the diagram stops settling.
Reported in eclipse-glsp/glsp#1724 with the node server and the VS Code integration.

Until eclipse-glsp/glsp-client#525 the collected bounds were only cleared on the path that
actually sends them, so an export or a failing bounds pass left them behind for the next run.
Rather than relying on that cleanup alone, the client now never names a client-only element in
a request to the server.

- Mark `GIssueMarker` and the dangling feedback edge as feedback elements. `feedbackFeature`
  already existed for that concept but was only ever set and never read
- Skip feedback elements and their routes when answering a server `RequestBoundsAction`. Local
  bounds requests keep reporting them, so decoration placement and ghost sizing are unaffected
- Reset the collected bounds when a hidden rendering starts rather than only when `postUpdate`
  completes, which the viewer never reaches when a view or the vdom patch throws
- Add `enableFeatures` to the model utilities. The feature set is established once per element
  type and shared by all its instances, so it copies before extending

Fixes eclipse-glsp/glsp#1725

#### How to test

The failure itself is no longer reproducible by hand on master, because #525 closed both known
leak paths. The remaining trigger needs a rendering that throws, which is covered by a spec, so
the automated tests are the primary evidence here.

```bash
pnpm install && pnpm build
pnpm test
```

To see the new specs fail without the fix:

```bash
# 2 failures: the marker filter and the interrupted hidden rendering
git stash push -- packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts \
                  packages/client/src/features/validation/issue-marker.ts
pnpm exec vitest run packages/client/src/features/bounds/glsp-hidden-bounds-updater.spec.ts
git stash pop

# 1 failure: the drawn feedback edge is not marked
git stash push -- packages/client/src/features/tools/edge-creation/dangling-edge-feedback.ts
pnpm exec vitest run packages/client/src/features/tools/edge-creation/dangling-edge-feedback.spec.ts
git stash pop
```

Manual check that nothing visible changed, since this only removes ids from what we send:

```bash
pnpm -C examples/workflow-standalone start
```

1. Click the validate button in the tool palette, markers appear on the task nodes
2. Markers are placed as before and their hover popups still work
3. Resize a marked node, the marker stays in place (local bounds requests still report it)
4. Draw an edge, the dangling feedback edge still follows the mouse and routes normally
5. Export with `Ctrl+Shift+E`, then delete a node, the model update still succeeds

- [x] Verified locally against the workflow example (node server)

#### Follow-ups

- The server still rejects the entire `computedBounds` action for a single unresolvable id, which
  is why one stale entry keeps a diagram from settling. Skipping unknown ids with a warning would
  degrade more gracefully. Deliberately out of scope here, worth its own ticket for the node and
  the Java server
- In the node server, `getOrThrow(index.get(...), 'Model element not found! ID: ...')` in
  `layout-util` never reports its message because `index.get` throws first. The Java server
  returns an `Optional` there and reports it as intended
- As noted in #525, sprotty's `HiddenBoundsUpdater` has the same cleanup-only-on-success shape, so
  plain-sprotty adopters are affected as well
- `ApplyTypeHintsCommand` extends the feature set in place, which is shared by all instances of an
  element type. It is harmless today because hints are per type, but `enableFeatures` now offers
  the safe alternative if that code is ever touched

#### Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Changes
 
 - [layout] Fix bounds computation error after diagram export [#525](https://github.com/eclipse-glsp/glsp-client/pull/525)
-- [layout] Keep client-side feedback elements, such as validation markers, out of the bounds reported to the server [#1725](https://github.com/eclipse-glsp/glsp/issues/1725)
+- [layout] Keep client-side feedback elements, such as validation markers, out of the bounds reported to the server [#531](https://github.com/eclipse-glsp/glsp-client/pull/531)
     - Adds the `enableFeatures` model utility to enable model features on a single element without modifying the feature set shared by its element type
 
 ### Potentially Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Changes
 
+- [layout] Fix bounds computation error after diagram export [#525](https://github.com/eclipse-glsp/glsp-client/pull/525)
+- [layout] Keep client-side feedback elements, such as validation markers, out of the bounds reported to the server [#1725](https://github.com/eclipse-glsp/glsp/issues/1725)
+    - Adds the `enableFeatures` model utility to enable model features on a single element without modifying the feature set shared by its element type
+
 ### Potentially Breaking Changes
 
 ## [v2.7.0 - 01/06/2026](https://github.com/eclipse-glsp/glsp-client/releases/tag/v2.7.0)
--- a/packages/client/src/features/bounds/glsp-hidden-bounds-updater.spec.ts
+++ b/packages/client/src/features/bounds/glsp-hidden-bounds-updater.spec.ts
@@ -289,6 +289,24 @@ describe('GLSPHiddenBoundsUpdater', () => {
         expect(computedRouteIds(dispatcher)).not.toContain(feedbackEdge.id);
     });
 
+    it('leaves out the routes an overridden calcElementRoute declines', () => {
+        const dispatcher = new RecordingActionDispatcher();
+        const updater = new (class extends TestHiddenBoundsUpdater {
+            protected override calcElementRoute(): undefined {
+                return undefined;
+            }
+        })(dispatcher);
+
+        const model = createRoot('node0');
+        addFeedbackEdge(model);
+
+        updater.renderHidden(model);
+        updater.postUpdate(LocalRequestBoundsAction.create(model));
+
+        // a local request reports feedback routes, so an empty result is down to the override alone
+        expect(computedBounds(dispatcher).routes).toBeUndefined();
+    });
+
     it('reports the route of a feedback edge for a local bounds request', () => {
         const dispatcher = new RecordingActionDispatcher();
         const updater = new TestHiddenBoundsUpdater(dispatcher);
--- a/packages/client/src/features/bounds/glsp-hidden-bounds-updater.spec.ts
+++ b/packages/client/src/features/bounds/glsp-hidden-bounds-updater.spec.ts
@@ -17,7 +17,10 @@ import {
     Action,
     Bounds,
     ComputedBoundsAction,
+    DefaultTypes,
     GModelElement,
+    GModelElementRegistration,
+    GModelFactory,
     GModelRoot,
     GModelRootSchema,
     GNode,
@@ -30,17 +33,24 @@ import {
     RequestBoundsAction,
     RequestExportAction,
     ResponseAction,
+    TYPES,
     Viewport,
     createFeatureSet
 } from '@eclipse-glsp/sprotty';
+import { Container } from 'inversify';
+import 'reflect-metadata';
 import { h } from 'snabbdom';
 import { describe, expect, it } from 'vitest';
 import { EditorContextService } from '../../base/editor-context-service';
 import { feedbackFeature } from '../../base/feedback/feedback-action-dispatcher';
 import { ServerAction } from '../../base/model/glsp-model-source';
+import { GModelRegistry } from '../../base/model/model-registry';
 import { GEdge, GGraph } from '../../model';
 import { enableFeatures } from '../../utils/gmodel-util';
-import { getOrCreateGIssueMarker } from '../validation/issue-marker';
+import { MARQUEE } from '../tools/marquee-selection/marquee-tool-feedback';
+import { MarqueeNode } from '../tools/marquee-selection/model';
+import { InsertIndicator } from '../tools/node-creation/insert-indicator';
+import { GIssueMarker, getOrCreateGIssueMarker } from '../validation/issue-marker';
 import { GLSPHiddenBoundsUpdater } from './glsp-hidden-bounds-updater';
 import { LocalRequestBoundsAction } from './local-bounds';
 
@@ -150,6 +160,25 @@ function createRoot(nodeId: string): GModelRoot {
     return root;
 }
 
+/**
+ * Wires the real model factory the way the client does, so a model built through it is subject to
+ * the same feature handling as in production, i.e. every element of a type shares the feature set
+ * derived from its registration.
+ */
+function createModelFactory(): GModelFactory {
+    const container = new Container();
+    const registrations: GModelElementRegistration[] = [
+        { type: DefaultTypes.GRAPH, constr: GGraph },
+        { type: DefaultTypes.NODE, constr: GNode },
+        { type: DefaultTypes.ISSUE_MARKER, constr: GIssueMarker },
+        { type: MARQUEE, constr: MarqueeNode }
+    ];
+    registrations.forEach(registration => container.bind(TYPES.SModelElementRegistration).toConstantValue(registration));
+    container.bind(TYPES.SModelRegistry).to(GModelRegistry).inSingletonScope();
+    container.bind(TYPES.IModelFactory).to(GModelFactory).inSingletonScope();
+    return container.get<GModelFactory>(TYPES.IModelFactory);
+}
+
 /** Stand-in for the dangling edge the edge-creation tool draws while the user is connecting. */
 function addFeedbackEdge(root: GModelRoot): GEdge {
     const edge = new GEdge();
@@ -201,6 +230,45 @@ describe('GLSPHiddenBoundsUpdater', () => {
         const model = createRoot('node0');
         const marker = getOrCreateGIssueMarker(model.children[0] as GNode);
         marker.issues.push({ message: 'invalid', severity: 'error' });
+
+        updater.renderHidden(model);
+        updater.postUpdate(serverBoundsRequest(model));
+
+        expect(computedBoundsIds(dispatcher)).toEqual(['node0']);
+    });
+
+    it('does not report issue marker bounds of a factory-built hidden model for a server bounds request', () => {
+        const dispatcher = new RecordingActionDispatcher();
+        const updater = new TestHiddenBoundsUpdater(dispatcher);
+
+        const liveModel = createRoot('node0');
+        const marker = getOrCreateGIssueMarker(liveModel.children[0] as GNode);
+        marker.issues.push({ message: 'invalid', severity: 'error' });
+
+        // the hidden rendering runs on a factory copy of the live root, and the marker registration
+        // contributes no feedback feature, so the marking only survives because the factory takes
+        // over the features of the element it is handed
+        const hiddenModel = createModelFactory().createRoot(liveModel);
+        const copiedMarker = (hiddenModel.children[0] as GParentElement).children.find(child => child instanceof GIssueMarker);
+        expect(copiedMarker?.hasFeature(feedbackFeature)).toBe(true);
+
+        updater.renderHidden(hiddenModel);
+        updater.postUpdate(serverBoundsRequest(hiddenModel));
+
+        expect(computedBoundsIds(dispatcher)).toEqual(['node0']);
+    });
+
+    it('does not report the bounds of the client-only elements the tools add for a server bounds request', () => {
+        const dispatcher = new RecordingActionDispatcher();
+        const updater = new TestHiddenBoundsUpdater(dispatcher);
+
+        const model = createRoot('node0');
+        const indicator = new InsertIndicator();
+        indicator.id = 'insert-indicator';
+        model.add(indicator);
+        // the marquee gets its features from its registration, just like in a running client
+        const marquee = createModelFactory().createElement({ type: MARQUEE, id: 'marquee' });
+        model.add(marquee);
 
         updater.renderHidden(model);
         updater.postUpdate(serverBoundsRequest(model));
--- a/packages/client/src/features/bounds/glsp-hidden-bounds-updater.spec.ts
+++ b/packages/client/src/features/bounds/glsp-hidden-bounds-updater.spec.ts
@@ -18,6 +18,7 @@ import {
     Bounds,
     ComputedBoundsAction,
     DefaultTypes,
+    EdgeRouterRegistry,
     GModelElement,
     GModelElementRegistration,
     GModelFactory,
@@ -47,6 +48,7 @@ import { ServerAction } from '../../base/model/glsp-model-source';
 import { GModelRegistry } from '../../base/model/model-registry';
 import { GEdge, GGraph } from '../../model';
 import { enableFeatures } from '../../utils/gmodel-util';
+import { routingModule } from '../routing/routing-module';
 import { MARQUEE } from '../tools/marquee-selection/marquee-tool-feedback';
 import { MarqueeNode } from '../tools/marquee-selection/model';
 import { InsertIndicator } from '../tools/node-creation/insert-indicator';
@@ -146,18 +148,45 @@ class TestHiddenBoundsUpdater extends GLSPHiddenBoundsUpdater {
     }
 }
 
+/** Bounds updater wired with the real routers, so an edge is routed as it is in a running client. */
+class RoutingTestHiddenBoundsUpdater extends TestHiddenBoundsUpdater {
+    protected override readonly edgeRouterRegistry = createRouterRegistry();
+}
+
 function createRoot(nodeId: string): GModelRoot {
     const root = new GGraph();
     root.id = 'root';
     root.type = 'graph';
     root.features = new Set<symbol>(GGraph.DEFAULT_FEATURES);
+    addNode(root, nodeId);
+    return root;
+}
+
+function addNode(root: GModelRoot, nodeId: string, position: Point = Point.ORIGIN): GNode {
     const node = new GNode();
     node.id = nodeId;
     node.type = 'node';
     node.features = new Set<symbol>(GNode.DEFAULT_FEATURES);
-    node.bounds = { x: 0, y: 0, width: 10, height: 10 };
+    node.bounds = { ...position, width: 10, height: 10 };
     root.add(node);
-    return root;
+    return node;
+}
+
+function addEdge(root: GModelRoot, sourceId: string, targetId: string): GEdge {
+    const edge = new GEdge();
+    edge.id = `${sourceId}-${targetId}`;
+    edge.type = 'edge';
+    edge.features = createFeatureSet(GEdge.DEFAULT_FEATURES);
+    edge.sourceId = sourceId;
+    edge.targetId = targetId;
+    root.add(edge);
+    return edge;
+}
+
+function createRouterRegistry(): EdgeRouterRegistry {
+    const container = new Container();
+    container.load(routingModule);
+    return container.get<EdgeRouterRegistry>(EdgeRouterRegistry);
 }
 
 /**
@@ -287,6 +316,34 @@ describe('GLSPHiddenBoundsUpdater', () => {
         updater.postUpdate(serverBoundsRequest(model));
 
         expect(computedRouteIds(dispatcher)).not.toContain(feedbackEdge.id);
+    });
+
+    it('leaves out the route of an edge the router cannot route', () => {
+        const dispatcher = new RecordingActionDispatcher();
+        const updater = new RoutingTestHiddenBoundsUpdater(dispatcher);
+
+        const model = createRoot('node0');
+        // the target is not part of the model, so the router reports an empty route
+        const edge = addEdge(model, 'node0', 'missing');
+
+        updater.renderHidden(model);
+        updater.postUpdate(serverBoundsRequest(model));
+
+        expect(computedRouteIds(dispatcher)).not.toContain(edge.id);
+    });
+
+    it('reports the route of an edge the router can route', () => {
+        const dispatcher = new RecordingActionDispatcher();
+        const updater = new RoutingTestHiddenBoundsUpdater(dispatcher);
+
+        const model = createRoot('node0');
+        addNode(model, 'node1', { x: 100, y: 100 });
+        const edge = addEdge(model, 'node0', 'node1');
+
+        updater.renderHidden(model);
+        updater.postUpdate(serverBoundsRequest(model));
+
+        expect(computedRouteIds(dispatcher)).toContain(edge.id);
     });
 
     it('leaves out the routes an overridden calcElementRoute declines', () => {
--- a/packages/client/src/features/bounds/glsp-hidden-bounds-updater.spec.ts
+++ b/packages/client/src/features/bounds/glsp-hidden-bounds-updater.spec.ts
@@ -30,13 +30,16 @@ import {
     RequestBoundsAction,
     RequestExportAction,
     ResponseAction,
-    Viewport
+    Viewport,
+    createFeatureSet
 } from '@eclipse-glsp/sprotty';
 import { h } from 'snabbdom';
 import { describe, expect, it } from 'vitest';
 import { EditorContextService } from '../../base/editor-context-service';
+import { feedbackFeature } from '../../base/feedback/feedback-action-dispatcher';
 import { ServerAction } from '../../base/model/glsp-model-source';
-import { GGraph } from '../../model';
+import { GEdge, GGraph } from '../../model';
+import { enableFeatures } from '../../utils/gmodel-util';
 import { getOrCreateGIssueMarker } from '../validation/issue-marker';
 import { GLSPHiddenBoundsUpdater } from './glsp-hidden-bounds-updater';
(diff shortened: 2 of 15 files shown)


packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts
```diff
@@ -112,7 +141,10 @@ export class GLSPHiddenBoundsUpdater extends HiddenBoundsUpdater {
                     layoutData.push({ elementId: element.id, layoutData: boundsData.layoutData });
                 }
             });
-            const routes = this.element2route.length === 0 ? undefined : this.element2route;
+            const relevantRoutes = skipFeedback
+                ? this.element2route.filter(route => !this.isFeedbackElementId(route.elementId))
```
tortmayr: The bounds filter tests the element directly, the route filter goes through `this.root?.index` instead, so it fails open and reports a route as non-feedback whenever the id does not resolve in the current root (`cleanUp` clears `this.root`, so this also depends on statement order in `postUpdate`). `decorate` has the routable element in hand, so the flag could be recorded there next to `element2route` and this lookup dropped.

packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts
```diff
@@ -54,18 +57,38 @@ export class GLSPHiddenBoundsUpdater extends HiddenBoundsUpdater {
...
+    /**
+     * Drops the data collected for the previous hidden rendering as soon as a new one starts.
+     * {@link postUpdate} cleans up on its way out, but it is not guaranteed to be reached at all:
+     * a failing view or vdom patch aborts the rendering before the viewer calls it, which would
+     * leave the collected bounds behind and report them with the next `ComputedBoundsAction`.
+     *
+     * Elements are decorated bottom-up, so the root cannot serve as the marker for a new rendering.
+     * Instead every element reports the root it belongs to, which changes with the rendering.
+     */
+    protected resetOnNewRendering(element: GModelElement): void {
```
tortmayr: This only fires when the root *instance* differs, which works because `RequestBoundsCommand`, `LocalRequestBoundsAction` and `RequestExportCommand` all build their hidden root via `modelFactory.createRoot` ([export path](https://github.com/eclipse-glsp/glsp-client/blob/e3f9011ed64dd4e5d6ecae1a3ef7d96a5f1aaeab/packages/client/src/features/export/request-export-command.ts#L45-L46)). That assumption is what makes the reset work, so it should be part of the doc comment.

packages/client/src/features/tools/edge-creation/dangling-edge-feedback.ts
```diff
@@ -162,6 +162,9 @@ export function drawFeedbackEdge(
 
     const feedbackEdge = context.modelFactory.createElement(edgeSchema);
     if (isRoutable(feedbackEdge)) {
+        // the edge is created from the adopter's own edge type, so it needs to be marked explicitly
+        // to be recognizable as client-side feedback
+        enableFeatures(feedbackEdge, feedbackFeature);
```
tortmayr, line 166: `drawMarquee` creates its client-only element the same way ([link](https://github.com/eclipse-glsp/glsp-client/blob/e3f9011ed64dd4e5d6ecae1a3ef7d96a5f1aaeab/packages/client/src/features/tools/marquee-selection/marquee-tool-feedback.ts#L121-L122)), and `InsertIndicator` ([link](https://github.com/eclipse-glsp/glsp-client/blob/e3f9011ed64dd4e5d6ecae1a3ef7d96a5f1aaeab/packages/client/src/features/tools/node-creation/insert-indicator.ts#L23)) plus the template elements of `AddTemplateElementsFeedbackCommand` ([link](https://github.com/eclipse-glsp/glsp-client/blob/e3f9011ed64dd4e5d6ecae1a3ef7d96a5f1aaeab/packages/client/src/features/element-template/add-template-element.ts#L67-L74)) are just as client-only, so the invariant only holds for markers and this edge. For the registered client-only types `feedbackFeature` can go straight into `DEFAULT_FEATURES` the way `FeedbackEdgeEnd` already does ([link](https://github.com/eclipse-glsp/glsp-client/blob/e3f9011ed64dd4e5d6ecae1a3ef7d96a5f1aaeab/packages/client/src/features/tools/edge-creation/dangling-edge-feedback.ts#L108)); templates are trees, and the filter only looks at the element itself, so their children need it too.

packages/client/src/utils/gmodel-util.ts
```diff
@@ -250,6 +250,28 @@ export function toggleCssClass(element: GModelElement, cssClass: string, toggle:
...
+ * @param element The element for which the features should be enabled.
+ * @param features The features to enable.
+ */
+export function enableFeatures(element: GModelElement, features: symbol[]): void;
+export function enableFeatures(element: GModelElement, ...features: symbol[]): void;
+export function enableFeatures(element: GModelElement, ...features: symbol[] | [symbol[]]): void {
+    const toEnable = Array.isArray(features[0]) ? features[0] : (features as symbol[]);
+    if (element.features === undefined) {
+        element.features = new Set<symbol>(toEnable);
+    } else if (element.features instanceof Set) {
```
tortmayr: If `features` is not a `Set` this silently does nothing and the element then quietly keeps flowing into server requests, i.e. the bug this PR fixes comes back without a trace. `ApplyTypeHintsCommand` uses the same guard ([link](https://github.com/eclipse-glsp/glsp-client/blob/e3f9011ed64dd4e5d6ecae1a3ef7d96a5f1aaeab/packages/client/src/features/hints/type-hint-provider.ts#L106)), but a skipped type hint is cosmetic. Falling back to a wrapping feature set (`{ has: f => toEnable.includes(f) || previous.has(f) }`) or at least logging would avoid the silent case.

packages/client/src/utils/gmodel-util.ts
```diff
@@ -250,6 +250,28 @@ export function toggleCssClass(element: GModelElement, cssClass: string, toggle:
...
+ * its instances, so the element is given a copy of it rather than having the shared one extended.
+ * Elements with a feature set that is not a `Set`, i.e. a hand-written `FeatureSet`, are left
+ * untouched as their features cannot be enumerated.
+ *
+ * @param element The element for which the features should be enabled.
+ * @param features The features to enable.
+ */
+export function enableFeatures(element: GModelElement, features: symbol[]): void;
+export function enableFeatures(element: GModelElement, ...features: symbol[]): void;
+export function enableFeatures(element: GModelElement, ...features: symbol[] | [symbol[]]): void {
```
tortmayr, line 266: The `Set` branch is copy-and-extend, which `createFeatureSet` already does and which is used that way in the repo ([link](https://github.com/eclipse-glsp/glsp-client/blob/e3f9011ed64dd4e5d6ecae1a3ef7d96a5f1aaeab/packages/client/src/features/test/layouter-test-util.ts#L49)): `element.features = createFeatureSet([...element.features], { enable: toEnable })`.

packages/client/src/features/bounds/glsp-hidden-bounds-updater.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+
+        expect(computedBoundsIds(dispatcher)).toEqual(['node0']);
+    });
+
+    it('does not report issue marker bounds for a server bounds request', () => {
+        const dispatcher = new RecordingActionDispatcher();
+        const updater = new TestHiddenBoundsUpdater(dispatcher);
+
+        const model = createRoot('node0');
+        const marker = getOrCreateGIssueMarker(model.children[0] as GNode);
```
tortmayr, line 260: In production the hidden model that carries the marker is a factory copy of the live root, and the marking only survives it because sprotty's `initializeElement` copies `features` by reference. The dangling-edge spec models the factory on purpose, this one builds the marker by hand, so nothing here would notice if that ever changed. One case that runs the marked root through `modelFactory.createRoot` would pin it.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### Changes
 
+- [layout] Fix bounds computation error after diagram export [#525](https://github.com/eclipse-glsp/glsp-client/pull/525)
+- [layout] Keep client-side feedback elements, such as validation markers, out of the bounds reported to the server [#1725](https://github.com/eclipse-glsp/glsp/issues/1725)
```
tortmayr: Every other entry links a glsp-client PR, this one links the glsp issue. Use [#531](https://github.com/eclipse-glsp/glsp-client/pull/531) and keep the issue reference in the PR body.

packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts
```diff
@@ -144,6 +176,20 @@ export class GLSPHiddenBoundsUpdater extends HiddenBoundsUpdater {
         this.root = undefined;
     }
 
+    /**
+     * Whether the given element only exists as client-side feedback, such as a validation marker.
```
tortmayr: method is pretty self explanatory. does not need an extensive comment

CHANGELOG.md
tortmayr: The changelog reference needs to be the pr not the issue.
Either update or remove. We  generate changelog entries before a release anyways. so manually adjusting is not necessary

tortmayr (changes requested):

packages/client/src/utils/gmodel-util.ts
```diff
@@ -387,13 +410,17 @@ export function calcElementAndRoutingPoints(element: GRoutableElement, routerReg
...
- * @returns The corresponding route for the given element.
+ * @returns The corresponding route for the given element, with at least a source and a target point.
  */
 export function calcElementAndRoute(element: GRoutableElement, routerRegistry?: EdgeRouterRegistry): ElementAndRoutingPoints {
     let route: Point[] | undefined = routerRegistry ? calcRoute(element, routerRegistry, ROUTE_KINDS) : undefined;
-    if (!route) {
+    // routers report that they cannot route an edge by returning an empty route, e.g. when an endpoint
+    // is not part of the model or has no anchor yet. an empty array is truthy, so the length is what
+    // distinguishes an unroutable edge from a routed one here
+    if (!route || route.length < 2) {
```
tortmayr: The two cases a router signals with `[]` need different answers. When it declines because an anchor is missing, source and target resolve and this fallback gives a sane coarse route. When it declines because an endpoint is not in the model, `element.source`/`element.target` are `undefined` and this produces `Point.ORIGIN` instead: I get `newRoutingPoints: [{x:100,y:100},{x:0,y:0}]` for an edge with an unresolvable target, and that edge id is one the server resolves, so it overwrites a real route with a point at the origin. Fabricating a route is worse than reporting none, so guard on `element.source && element.target` and leave the edge out otherwise (which `calcElementRoute` now supports).

Also, this is not part of #1725 and has no linked issue. It is a behaviour change on a public utility with three other call sites, so it would be easier to reason about in its own PR.

packages/client/src/features/tools/edge-creation/dangling-edge-feedback.ts
```diff
@@ -162,6 +162,9 @@ export function drawFeedbackEdge(
...
         edgeEnd.feedbackEdge = feedbackEdge;
         root.add(edgeEnd);
         root.add(feedbackEdge);
```
tortmayr, line 169: One more instance of the same thing: `drawFeedbackEdgeSource` in the edge-edit tool builds the identical client-only edge, with the same `feedbackEdgeId(root)` and a `FeedbackEdgeEnd` as its endpoint, and is still unmarked ([link](https://github.com/eclipse-glsp/glsp-client/blob/e3f9011ed64dd4e5d6ecae1a3ef7d96a5f1aaeab/packages/client/src/features/tools/edge-edit/edge-edit-tool-feedback.ts#L376-L380)). Its route goes to the server on reconnect.

packages/client/src/features/validation/issue-marker.ts
```diff
@@ -25,11 +25,14 @@ import {
...
 } from '@eclipse-glsp/sprotty';
+import { feedbackFeature } from '../../base/feedback/feedback-action-dispatcher';
 
 export class GIssueMarker extends SIssueMarkerImpl implements Projectable {
     constructor() {
         super();
-        this.features = new Set<symbol>(GDecoration.DEFAULT_FEATURES);
+        // markers are established as client-side feedback and are unknown to the server,
+        // so they are marked as such to keep them out of server-bound requests
+        this.features = new Set<symbol>([...GDecoration.DEFAULT_FEATURES, feedbackFeature]);
```
tortmayr: This assignment does not survive the registration. `GIssueMarker` is registered via `configureModelElement` ([link](https://github.com/eclipse-glsp/glsp-client/blob/e3f9011ed64dd4e5d6ecae1a3ef7d96a5f1aaeab/packages/client/src/views/base-view-module.ts#L91)) and `GModelRegistry.getDefaultFeatures` always returns a non-empty array ([link](https://github.com/eclipse-glsp/glsp-client/blob/e3f9011ed64dd4e5d6ecae1a3ef7d96a5f1aaeab/packages/client/src/base/model/model-registry.ts#L44-L51)), so sprotty's registry factory overwrites `element.features` right after the constructor ran. I checked: a marker built from a schema through the model factory has no `feedbackFeature`. It only works today because `getOrCreateGIssueMarker` uses `new GIssueMarker()` and the hidden root is a factory copy that takes the features over.

`static override readonly DEFAULT_FEATURES = [...GDecoration.DEFAULT_FEATURES, feedbackFeature]` holds on both paths and matches what you did for `InsertIndicator` and `MarqueeNode` in this PR. If the reason against it is that a server may send `marker` elements itself, say so here, because that is not obvious.

packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts
```diff
@@ -54,18 +58,66 @@ export class GLSPHiddenBoundsUpdater extends HiddenBoundsUpdater {
...
 
+    /**
+     * Ids of the routable elements of {@link element2route} that only exist as client-side feedback.
+     * Recorded while decorating, where the element itself is at hand, because the routes outlive the
+     * root they were collected from.
+     */
+    protected feedbackRouteIds = new Set<string>();
+
+    /** Root of the hidden rendering currently being collected, used to detect the start of the next one. */
+    protected collectingForRoot?: GModelRoot;
```
tortmayr: This keeps the last hidden root alive after the rendering is done, which is exactly what `cleanUp` avoids by dropping `this.root`. Hidden roots are throwaway factory copies of the whole model, so a `WeakRef` (or clearing it in `cleanUp` and comparing against `undefined`) keeps the reset without holding on to the tree.

packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts
```diff
@@ -54,18 +58,66 @@ export class GLSPHiddenBoundsUpdater extends HiddenBoundsUpdater {
...
+     * {@link ComputedBoundsAction} of this rendering.
+     *
+     * Override to substitute or to skip a route. Skipping is lossless as long as a later rendering
+     * reports it, which is what makes this the place to opt out of routing an element whose geometry
+     * has not been measured yet: the first hidden rendering of a model still carries the placeholder
+     * size `Dimension.EMPTY`, so a router either declines to route at all or works from geometry that
+     * the visible rendering immediately supersedes. `Dimension.isValid` on the endpoint sizes detects
+     * that state.
+     */
+    protected calcElementRoute(element: GRoutableElement): ElementAndRoutingPoints | undefined {
```
tortmayr: Nothing in the PR overrides this, and the doc spends most of its length on a `Dimension.isValid` placeholder-size scenario that is not implemented anywhere in the repo. Either drop the seam (`decorate` can call `calcElementAndRoute` as before) or keep it with a one-liner. Same point as my earlier comment on `isFeedbackElement`: short docs, and no docs for hypotheticals. `resetOnNewRendering` below has the same shape, the first paragraph says it, the second is commentary.

packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts
```diff
@@ -141,9 +202,14 @@ export class GLSPHiddenBoundsUpdater extends HiddenBoundsUpdater {
     protected cleanUp(): void {
         this.getElement2BoundsData().clear();
         this.element2route = [];
+        this.feedbackRouteIds.clear();
         this.root = undefined;
     }
 
+    protected isFeedbackElement(element: GModelElement): boolean {
+        return element.hasFeature(feedbackFeature);
```
tortmayr: `feedbackFeature` goes from a marker nobody read to "the server cannot resolve this element's id" with this PR. That contract belongs at the declaration ([link](https://github.com/eclipse-glsp/glsp-client/blob/e3f9011ed64dd4e5d6ecae1a3ef7d96a5f1aaeab/packages/client/src/base/feedback/feedback-action-dispatcher.ts#L37)), which has no doc at all. Adopters with their own client-only elements will look there, not here.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### Changes
 
+- [layout] Fix bounds computation error after diagram export [#525](https://github.com/eclipse-glsp/glsp-client/pull/525)
+- [layout] Keep client-side feedback elements, such as validation markers, out of the bounds reported to the server [#531](https://github.com/eclipse-glsp/glsp-client/pull/531)
+    - Adds the `enableFeatures` model utility to enable model features on a single element without modifying the feature set shared by its element type
```
tortmayr: The `calcElementAndRoute` change is missing here. It is exported API with three other call sites and it now reports a source/target fallback where it used to pass an empty route through, so adopters should see it.

packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts
martin-fleck-at: Done. The flag is now recorded in `decorate` right next to the route, so the lookup through the root index is gone.

packages/client/src/features/tools/edge-creation/dangling-edge-feedback.ts
martin-fleck-at: Good catch. `InsertIndicator` and `MarqueeNode` now carry the feature via their `DEFAULT_FEATURES`, and template elements are marked together with their children.

packages/client/src/utils/gmodel-util.ts
martin-fleck-at: Changed to use `createFeatureSet` as suggested.

packages/client/src/features/bounds/glsp-hidden-bounds-updater.spec.ts
martin-fleck-at: Added a case that runs the marked model through the real model factory and checks the marking survives the copy. It fails if the marker is no longer marked as feedback.

packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts
martin-fleck-at: Removed. The reason is already stated at the call site.

CHANGELOG.md
martin-fleck-at: Updated to point at the PR instead of the issue. Happy to drop the entry entirely if you prefer, since the entries get generated before a release anyway.

martin-fleck-at:
@tortmayr Thank you very much, I addressed your review comments in a follow-up commit. 

I also added three more commits with minor things:
- `43108cd` treats a route with less than two points as a missing route. This is a separate cause of the same symptom: routers return an empty route when they cannot route an edge, and the truthiness check let that through, so the server rejected the whole computedBounds action. Probably worth a look on its own.
- `0017a1c` makes the per-element route overridable, so a subclass can substitute or skip a route instead of correcting the collected list afterwards.
- `a43d859` is unrelated housekeeping: prettier formatting on SECURITY.md, which was failing format:check on master.

Happy to split any of them out if you would rather keep this PR narrow.

tortmayr (changes requested):
Thanks for addressing the review comments. 
I found a couple minor new issues but apart from that code looks good to me

packages/client/src/features/tools/edge-creation/dangling-edge-feedback.ts
martin-fleck-at: Done, `drawFeedbackEdgeSource` marks the edge the same way now, with a spec mirroring the
edge-creation one.

packages/client/src/features/validation/issue-marker.ts
martin-fleck-at: Confirmed, the registration overwrites `features` right after the constructor ran, so the
assignment only held for `new GIssueMarker()`.

I did not go with `DEFAULT_FEATURES` though, because `DefaultTypes.ISSUE_MARKER` is a type a
server can send as part of the model, the node server has a `GIssueMarker` graph element for it,
and those bounds do need to be reported back. Instead the marking now happens in
`getOrCreateGIssueMarker`, where client-side validation creates the marker, the same way
`drawFeedbackEdge` marks its edge. That keeps the distinction explicit rather than resting on
`new` versus factory and on `initializeElement` copying `features` by reference.

Behaviour is unchanged on both paths, so the existing specs still cover it, but the intent no
longer depends on how the element was built. Happy to switch to `DEFAULT_FEATURES` if you would
rather have all markers count as feedback.

packages/client/src/features/bounds/glsp-hidden-bounds-updater.ts
martin-fleck-at: Done, it is a `WeakRef` now and `cleanUp` drops it as well, so a finished rendering no longer
pins the hidden root.

That also leaves a cleaner invariant: after a rendering that reached `postUpdate` the field is
`undefined` and the next rendering resets unconditionally, so the identity comparison is only
load-bearing for a rendering that died before `postUpdate`, which is exactly the case the reset
exists to cover.

packages/client/src/utils/gmodel-util.ts
martin-fleck-at: Agreed on both counts, and I dropped the change from this PR rather than guarding it.

Meanwhile eclipse-glsp/glsp-server-node#149 landed for eclipse-glsp/glsp#1727: the server now
reports a route of fewer than two points as not applicable and logs it at debug, since a client
may have nothing to report for an unmeasured edge yet. So an empty route no longer costs the
whole action, while the fabricated route you found does resolve and does have two points, which
makes it strictly worse than reporting none. `calcElementAndRoute` is back to its state on
master, specs included.

What is left on the client is not to send such a route in the first place, and that is the
concern of the bounds pass alone, so `calcElementRoute` now leaves out a route of fewer than two
points. The new spec fails without the skip with
`expected [ 'node0-missing' ] to not include 'node0-missing'`, i.e. your scenario.

The utility itself still hands `newRoutingPoints: []` to its other two call sites for an
unroutable edge, unchanged from master. The Java server also still throws `Invalid Route!` there,
which is the open half of #1727. Happy to take that on in its own PR.

martin-fleck-at:
@tortmayr Thanks, addressed in `be08ccc`. Two of the four are straight fixes, two come with a note:

- The reconnect feedback edge of the edge-edit tool is marked now, with a spec mirroring the edge-creation one
- `collectingForRoot` is a `WeakRef` and `cleanUp` drops it, so a finished rendering no longer pins the hidden root
- The marker is marked in `getOrCreateGIssueMarker` instead of via `DEFAULT_FEATURES`, because a server can send `marker` elements itself and those bounds have to be reported back. Details in the thread, easy to switch if you disagree
- The route change is reverted out of this PR. eclipse-glsp/glsp-server-node#149 landed for eclipse-glsp/glsp#1727 and now treats a route of fewer than two points as not applicable, so the fabricated route was the only remaining problem. Not sending one is the concern of the bounds pass, so `calcElementRoute` skips it there

The follow-ups section above is outdated in one point: the server-side leniency I listed there is eclipse-glsp/glsp#1727, with the node server half merged and the Java server still open.

tortmayr (approved):
LGTM!

